Issue #8, opened 2019-05-02
Set up node core


slifty:
We need to initialize the project with the basic node / yarn stack, and update the README to help devs know how to set up and run the project.

PR #12, opened 2019-05-02
Set up node

slifty:
This sets up our baseline node structure, including linting, bundling,
and transpiling.

As is the way, we're using airbnb's linter and not using semicolons.

This does NOT set up things like sequelize or react, as those should be
included when we're working on tasks that would actually need them.

Resolves Issue #8


slifty:
@reefdog I installed because it yelled at me when I `yarn lint`ed and assumed it was a plugin required by airbnb style.  Maybe it was because I had that copypasta preset, I'll check...

package.json
```diff
@@ -0,0 +1,33 @@
...
+    "@babel/core": "^7.4.4",
+    "@babel/node": "^7.2.2",
+    "@babel/preset-env": "^7.4.4",
+    "babel-eslint": "^10.0.1",
+    "concurrently": "^4.1.0",
+    "eslint": "^5.16.0",
+    "eslint-config-airbnb": "^17.1.0",
+    "eslint-plugin-import": "^2.17.2",
+    "eslint-plugin-jsx-a11y": "^6.2.1",
+    "eslint-plugin-react": "^7.12.4",
```
reefdog: Was thinking about these two also (JSX and React) but also these are the smallest of nitpicks. If we're going to add React later, then fine fine.

slifty: Right -- those are required by the airbnb style, so no real choice.

PR #38, opened 2019-05-09
Add sequelize to project

slifty:
This PR sets up sequelize, pg, and dotenv as project dependencies, sets up the initial configuration, and creates the directory skeleton.

## After merging this you will need to:

1. Set up a .env file that contains:
   * DATABASE_URL_PRODUCTION :: The full database url to be used in production.
   * DATABASE_URL_DEVELOPMENT :: The full database url to be used in development.
   * DATABASE_URL_TEST :: The full database url to be used in testing.
2. Run `yarn install`

This resolves #17

Issue #20, opened 2019-05-02
Set up common scrape tooling


reefdog:
We know we're going to want some common tooling shared across the scrapers, irrespective of site. Create a Hello World scraper that demonstrates:

- [x] Fetching a specific URL
- [x] Navigating the DOM looking for a certain type of link and following it
- [x] Extracting the target contents of the page
- [x] Getting the content in a state ready to be passed to the parsing step (ClaimBuster API, etc.) in an error-recoverable way (e.g., we shouldn't lose the data if the API call doesn't succeed)

slifty:
From what I have read, it looks like [cheerio](https://www.npmjs.com/package/cheerio) is the standard package used for navigating DOM / scraping in Node.

I'm spending some time right now looking at potential architecture for the general tooling; we should keep in mind that the nature of our scrapes make it so the "general" tooling is probably going to be third party libraries like cheerio. For instance, I think the CNN transcript crawl logic is going to be unique to CNN.

Basically, since we're very targeted in our scraping and completion is so important, I think the "abstract" parts of this scrape is really more at a pipeline and error handling level.

slifty:
I am looking to use a queue manager to keep track of crawls and scrapes -- realizing maybe this should be an issue in itself, but I'll just lump it in this anyway.

Looking around it seems like [Kue](https://www.npmjs.com/package/kue) used to be the go to lib for node, but it hasn't been updated in two years so instead I'll be looking at [bull](https://www.npmjs.com/package/bull).  Will circle back if I end up finding something else entirely.

slifty:
I'm not QUITE ready to close this because I would like to create something that more gracefully handles HTTP errors etc.  Maybe that should happen as a separate issue (since for now I think it probably handles it reasonably well actually; in that it would write an error to the log and crash that specific scrape process which is basically what would happen anyway)

@reefdog what do you think?

reefdog:
I say close this and open a new enhancement issue for that.

Issue #2, opened 2019-04-26
Diagram technical architecture


reefdog:
Using what we know about the project, create a technical architecture diagram we can use for  gap analysis of the [existing codebase](https://github.com/oneroyalace/Tech-Check-Alerts) and future development.

reefdog:
Quick v1.0 version of the technical architecture diagram. This will get more detail.

<img width="583" alt="diagram" src="https://user-images.githubusercontent.com/4731/56840787-1bd28f80-684f-11e9-8ad6-83b2e7b67eb6.png">

slifty:
Some other nodes to add:

1) Claimbuster, which is a third party API.
2) We may want to wrap the DB in an API layer, rather than having multiple applications work directly with the same database.

slifty:
Capturing the outcome and deliberations of our discussion this morning!

## The Approaches
We were trying to choose between two approaches:

### Approach 1: Monolithic application

The first option would be to put the subscription service in the same project as the scrapers.  This would have the benefits of:

1. Simpler architecture; easier to maintain and set up.  You don't have multiple CI and automated deployment processes going on for each project.
2. Common database and model / schema files.  Both projects that interact with the database would be able to leverage the same models.  Avoiding redundancy

### Approach 2: Distinct services

Decoupling the code that generates content from the code that uses the content by putting them in separate projects and deploying them separately.

This would mean breaking the core claim database into its own service that is used by both pieces of the application.  The side applications might have their own databases to manage their parts of the world (e.g. email lists)

This would have the benefit of:

1. Ability to scale each piece of the project separately.
2. More flexibility down the line for creating new types of "clients" that use this data.

## The Decision

Since this project isn't too complex right now, we're going with Approach #1 and having a single project that handles everything under one roof.  This does not prevent us from moving into a more complex architecture in future if scraping starts to become a much more involved process / the data starts becoming too great, or if we start building out more robust subscription management and consumer interfaces.

This also doesn't prevent us from incorporating talking point tracker data at some point in the future (likely by creating a scraper module that scrapes TPT)

@reefdog is going to update the diagram to reflect these decisions!

reefdog:
Here's largely what we discussed:

1. **Node.js scrapers** fetch data from target sites
2. **Node.js parsers** pass data to an external ClaimBuster API and potentially additional to additional services (like a Python spaCy service we spin up). Technically, this will be coupled to the scraper, but I'm representing it separately for clarity.
[...]

PR #40, opened 2019-05-10
Add scraper tooling

slifty:
NOTE: it looks like this has the commits from #38; lets merge that and rebase before reviewing this.

This PR is the first step towards creating some common scraper tooling (#20)

Some notes about the design choices in this PR:

1. This sets up an AbstractCrawler and an example implemented crawler.
2. Crawlers pull out URLs to be scraped, scrapers pull out claims from URLs.
3. When we eventually build the scrapers we should consider having AbstractCrawler extend AbstractScraper. A lot of the general error handling, etc, will be common for both the crawlers and the scrapers.
4. This triggers the `HelloWorldCrawler` directly from `index.js`.  This will be replaced with something like an [EventEmitter](https://nodejs.org/api/events.html) eventually.
5. We may eventually want to have the crawler support recursive crawls (e.g. multi step crawls) to enable things like like pagination, etc.

### DIFF
--- a/src/server/workers/crawlers/AbstractCrawler.js
+++ b/src/server/workers/crawlers/AbstractCrawler.js
@@ -1,0 +1,36 @@
+import rp from 'request-promise'
+
+class AbstractCrawler {
+  constructor(crawlUrl) {
+    this.crawlUrl = crawlUrl
+  }
+
+  /**
+   * Process the result of an HTTP request and identify the urls
+   * to pass into the scraper pipeline.
+   *
+   * Each crawler that extends AbstractCrawler needs to implement its own
+   * crawlHandler method.
+   *
+   * @param string htmlString The HTML or JSON that came from the HTTP request
+   * @return [String] the list of URLs that should be scraped
+   */
+  // eslint-disable-next-line no-unused-vars
+  crawlHandler = (responseString) => {
+    throw new Error('You wrote a new crawler but forgot to define the crawlHandler.')
+  }
+
+  /**
+   * Runs the crawler
+   * @return Promise a promise that returns the list of URLs that came fromthe crawl handler
+   */
+  async run() {
+    return rp(this.crawlUrl)
+      .then(this.crawlHandler)
+      .catch((err) => {
+        console.log(err)
+      })
+  }
+}
+
+export default AbstractCrawler
--- a/.babelrc
+++ b/.babelrc
@@ -3,5 +3,6 @@
     "@babel/preset-env"
   ],
   "plugins": [
+    "@babel/plugin-proposal-class-properties"
   ]
 }
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "devDependencies": {
     "@babel/core": "^7.4.4",
     "@babel/node": "^7.2.2",
+    "@babel/plugin-proposal-class-properties": "^7.4.4",
     "@babel/preset-env": "^7.4.4",
     "babel-eslint": "^10.0.1",
     "concurrently": "^4.1.0",
@@ -31,8 +32,11 @@
     "parcel": "^1.12.3"
   },
   "dependencies": {
+    "cheerio": "^1.0.0-rc.3",
     "dotenv": "^8.0.0",
     "pg": "^7.10.0",
+    "request": "^2.88.0",
+    "request-promise": "^4.2.4",
     "sequelize": "^5.8.5"
   }
 }
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -1,0 +1,4 @@
+import HelloWorldCrawler from './workers/crawlers/HelloWorldCrawler'
+
+const crawler = new HelloWorldCrawler()
+crawler.run()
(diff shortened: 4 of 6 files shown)


src/server/workers/crawlers/AbstractCrawler.js
```diff
@@ -0,0 +1,36 @@
+import rp from 'request-promise'
+
+class AbstractCrawler {
+  constructor(crawlUrl) {
+    this.crawlUrl = crawlUrl
+  }
+
+  /**
+   * Processes the result of an HTTP request and identify the urls
```
reefdog: Can you unify the verbs here? "Processes/identifies" or "Process/identify"

src/server/workers/crawlers/AbstractCrawler.js
```diff
@@ -0,0 +1,36 @@
...
+   * to pass into the scraper pipeline.
+   *
+   * Each crawler that extends AbstractCrawler needs to impelment its own
```
reefdog: > impelment

I promise the only reason all my reviews today are grammar/spelling are because everything else looks good. And that's a Good Thing!

src/server/workers/crawlers/AbstractCrawler.js
```diff
@@ -0,0 +1,36 @@
...
+   * crawlHandler method.
+   *
+   * @param string htmlString The HTML or JSON that came from the HTTP request
+   * @return [String] the list of URLs that should be scraped
+   */
+  crawlHandler = (responseString) => {
+    console.log(responseString)
+    throw new Error('You wrote a new crawler but forgot to define the crawlHandler.')
```
reefdog: Just a note, I think this is a great way to handle requiring `crawlHandler` be defined by the extended class. I've somehow not seen this pattern before. 🙌

src/server/workers/crawlers/AbstractCrawler.js
```diff
@@ -0,0 +1,36 @@
...
+   * to pass into the scraper pipeline.
+   *
+   * Each crawler that extends AbstractCrawler needs to impelment its own
+   * crawlHandler method.
+   *
+   * @param string htmlString The HTML or JSON that came from the HTTP request
+   * @return [String] the list of URLs that should be scraped
+   */
+  crawlHandler = (responseString) => {
+    console.log(responseString)
```
reefdog: …except should we remove this console.log?

src/server/workers/crawlers/AbstractCrawler.js
```diff
@@ -0,0 +1,36 @@
...
+    throw new Error('You wrote a new crawler but forgot to define the crawlHandler.')
+  }
+
+  /**
+   * Runs the crawler
+   * @return {[type]} [description]
```
reefdog: Replace boilerplate with actual `@return` documentation?

src/server/workers/crawlers/AbstractCrawler.js
slifty: What about Procidentifyessy

src/server/workers/crawlers/AbstractCrawler.js
slifty: I admit: I added that because eslint errors if you don't use the variable.  Which is a bad reason so ty for calling it out.

My options are:

1. Remove the parameter (but I want to keep it because it is defining an abstract function and I want folks to know the signature)
2. Add a `// eslint-disable-next-line` somewhere
3. Figure out some other lame hackey reason for using it
4. Laugh maniacally as the linter breaks for the rest of the project

I'll go with #2 but I'll still feel bad about myself.

slifty:
Awesome -- yeah this is not intended to close out issue 20; It's just intended as a first step.

I want to nail down the event emitters and error handling before we can consider calling it a wrap

src/server/workers/crawlers/AbstractCrawler.js
reefdog: If it wasn't too late I'd merge-block this with a change request just because of your self-👍.

src/server/workers/crawlers/AbstractCrawler.js
reefdog: Ahh, that's a sensible reason. I think stubs are valuable, and if the Bad Consequence is a complaining linter, then disabling the linter (with a note as to why) is the most legible route.

slifty: I did the first part but not the second, nuts!!

reefdog: Ah yes, let's document anytime we disable the linter. (Ran into that recently on another project and was equally confused as to whether it was temporary or permanent.)

slifty: Let me just use my git timetravel-commit alias...

reefdog: `git timetravel-commit` sounds not unlike [`git-blame-someone-else`](https://github.com/jayphelps/git-blame-someone-else).

PR #42, opened 2019-05-16
Add models and migrations for claims and speakers

### LINKED ISSUE #14: Design data schema for claims
We need to store claims.

There's an existing [schema](https://github.com/oneroyalace/Tech-Check-Alerts/blob/master/db_create.sql) we can reference, but we should assess how much of that schema we want to directly replicate. For instance, can we normalize claim data into a common set of columns, with site-specific data in a JSONB column (or in a separate metadata table)?

reefdog:
Here we’ve got Sequelize models and migrations for claims and speakers, as well as a seed of each. After merging this, you’ll want to run `yarn migrate` and `sequelize db:seed:all` (the latter only if you actually want the seeds in your database, so, *not in production*).

By default, Sequelize camelCases tables and columns for consistency with the models and model attributes themselves, but we like to follow snake-case for SQL and camelCase for JavaScript. So there’s also some stitching here to make that mapping work properly. Just be aware that when referencing SQL field names (such as in migrations and seed files), you’ll want to use the snake-cased names (e.g. `created_at`), while using camelCase (`createdAt`) when referencing Sequelize/JavaScript model names/attributes. It can be weird.

Closes #14

### DIFF
--- a/src/server/migrations/20190516164909-create-claim.js
+++ b/src/server/migrations/20190516164909-create-claim.js
@@ -1,0 +1,31 @@
+module.exports = {
+  up: (queryInterface, Sequelize) => queryInterface.createTable('claims', {
+    id: {
+      allowNull: false,
+      autoIncrement: true,
+      primaryKey: true,
+      type: Sequelize.INTEGER,
+    },
+    content: {
+      type: Sequelize.TEXT,
+    },
+    claimed_at: {
+      type: Sequelize.DATE,
+    },
+    source_url: {
+      type: Sequelize.STRING(1024),
+    },
+    claim_buster_score: {
+      type: Sequelize.FLOAT,
+    },
+    created_at: {
+      allowNull: false,
+      type: Sequelize.DATE,
+    },
+    updated_at: {
+      allowNull: false,
+      type: Sequelize.DATE,
+    },
+  }),
+  down: queryInterface => queryInterface.dropTable('claims'),
+}
--- a/src/server/models/index.js
+++ b/src/server/models/index.js
@@ -1,0 +1,47 @@
+import fs from 'fs'
+import path from 'path'
+import Sequelize from 'sequelize'
+import config from '../../../config/sequelize-config'
+
+const env = process.env.NODE_ENV || 'development'
+const envConfig = config[env]
+
+const db = {}
+const sequelize = new Sequelize(envConfig.url, envConfig)
+
+const getModelFiles = () => {
+  const isFileVisible = file => file.indexOf('.') !== 0
+  const isFileDifferentFromThisFile = file => file !== path.basename(__filename)
+  const isFileJavaScript = file => file.slice(-3) === '.js'
+
+  return fs
+    .readdirSync(__dirname)
+    .filter(file => (
+      isFileVisible(file) && isFileDifferentFromThisFile(file) && isFileJavaScript(file)))
+}
+
+// Uses the `db` and `sequelize` variables defined above thanks to lexical scoping
+const registerModels = (modelFiles) => {
+  modelFiles.forEach((modelFile) => {
+    const model = sequelize.import(path.join(__dirname, modelFile))
+    db[model.name] = model
+  })
+}
+
+// Uses the `db` variable defined above thanks to lexical scoping
+const registerModelAssociations = () => {
+  Object.keys(db).forEach((modelName) => {
+    if (db[modelName].associate) {
+      db[modelName].associate(db)
+    }
+  })
+}
+
+const modelFiles = getModelFiles()
+registerModels(modelFiles)
+registerModelAssociations()
+
+db.sequelize = sequelize
+db.Sequelize = Sequelize
+
+module.exports = db
--- a/src/server/seeders/20190516191610-demo-speakers-and-claims.js
+++ b/src/server/seeders/20190516191610-demo-speakers-and-claims.js
@@ -1,0 +1,44 @@
+module.exports = {
+  up: async (queryInterface) => {
+    // Generate speakers first, so we have an ID to associate with claims.
+    await queryInterface.bulkInsert('speakers', [
+      {
+        full_name: '[Testing] Ray Boyd',
+        created_at: (new Date()).toISOString(),
+        updated_at: (new Date()).toISOString(),
+      },
+    ], {})
+
+    // Nab the latest speaker ID.
+    const speakers = await queryInterface.sequelize.query("SELECT id FROM speakers WHERE full_name LIKE '[Testing]%' ORDER BY id DESC LIMIT 1;")
+
+    // Now, generate our claims and associate with the speaker above.
+    return queryInterface.bulkInsert('claims', [
+      {
+        content: '[Testing] The human head weighs eight pounds.',
+        source_url: 'https://www.imdb.com/title/tt0116695/quotes/qt0389281',
+        claim_buster_score: 0.05,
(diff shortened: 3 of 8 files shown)


src/server/migrations/20190516164909-create-claim.js
```diff
@@ -0,0 +1,31 @@
...
+      type: Sequelize.TEXT,
+    },
+    timestamp: {
+      type: Sequelize.DATE,
+    },
+    source_url: {
+      type: Sequelize.STRING(1024),
+    },
+    claim_buster_score: {
+      type: Sequelize.DOUBLE,
```
slifty: Not a major deal but will claim buster scores need the precision of a DOUBLE?  Might FLOAT do the trick well enough (at half the size)?

src/server/models/index.js
```diff
@@ -0,0 +1,30 @@
...
+const basename = path.basename(__filename)
+const env = process.env.NODE_ENV || 'development'
+const envConfig = config[env]
+
+const db = {}
+const sequelize = new Sequelize(envConfig.url, envConfig)
+
+fs
+  .readdirSync(__dirname)
+  .filter(file => (file.indexOf('.') !== 0) && (file !== basename) && (file.slice(-3) === '.js'))
```
slifty: I know this is boilerplate but it would be good to put each of these checks in named methods to make it clearer what exactly you're filtering for.  (e.g. "isFileNotHidden" "isFileNotThis" and "isFileJs")

Side note then you can put basename (is that the right name for that?) inside of the isFileNotThis method

src/server/models/index.js
```diff
@@ -0,0 +1,30 @@
...
+  .forEach((file) => {
```
slifty: Consider breaking the chain and making a named method for the model registration (e.g. `registerModels(modelPaths)`)

It might also be best practice to either pass in everything needed by the method (I believe JS passes objects and arrays by reference, so you should still be able to modify the array / object directly).

src/server/models/index.js
```diff
@@ -0,0 +1,30 @@
...
+    const model = sequelize.import(path.join(__dirname, file))
+    db[model.name] = model
+  })
+
+Object.keys(db).forEach((modelName) => {
```
slifty: Same here -- putting this in a named method like `registerModelAssociations()` might help with legibility.

src/server/seeders/20190516191610-demo-speakers-and-claims.js
```diff
@@ -0,0 +1,34 @@
...
+      },
+    ], {})
+
+    // Nab the latest speaker ID.
+    const speakers = await queryInterface.sequelize.query('SELECT id FROM speakers ORDER BY id DESC LIMIT 1;')
+
+    // Now, generate our claims and associate with the speaker above.
+    return queryInterface.bulkInsert('claims', [
+      {
+        content: 'The human head weighs eight pounds.',
```
slifty: I'm gonna need proof for this.

src/server/models/index.js
slifty: Lol re-reading: 'isFileNotThis' is a terrible name.

src/server/migrations/20190516164909-create-claim.js
```diff
@@ -0,0 +1,31 @@
...
+    id: {
+      allowNull: false,
+      autoIncrement: true,
+      primaryKey: true,
+      type: Sequelize.INTEGER,
+    },
+    content: {
+      type: Sequelize.TEXT,
+    },
+    timestamp: {
```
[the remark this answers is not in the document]
reefdog: It's the time the claim was made, which we can parse from all the platforms. For TV transcripts, the granularity will probably just be the air date (or, at best, the program start time), not claim time. But for Twitter/FB posts we can grab the actual post time.

I'm open to renaming `timestamp` to something more like `claimedAt` or something, but I do want to keep it distinct from the auto-managed `createdAt`/`updatedAt` columns.

Of course, when choosing new claims to send, we should query by `createdAt`. But if we display to users when the claim was made, we'd want to use `timestamp` (or `claimedAt`).

What do you think?

src/server/migrations/20190516164909-create-claim.js
reefdog: Probably? The current score is [stored as a `DOUBLE`](https://github.com/TechAndCheck/tech-and-check-alerts-2018/blob/ab1bb599319f034dbe300d2890b1638df764d89e/db_create.sql#L3), so I just followed suit. @oneroyalace can we use `FLOAT` instead for ClaimBuster score?

src/server/seeders/20190516191610-demo-speakers-and-claims.js
reefdog: [[citation]](https://getyarn.io/yarn-clip/c43c407e-da27-4695-a7cd-cc81a400b516)

src/server/migrations/20190516164909-create-claim.js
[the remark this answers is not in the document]
slifty: claimedAt sounds good to me!

src/server/migrations/20190516164909-create-claim.js
slifty: @reefdog in that case I'm guessing it was a casual choice and we can safely use FLOAT (@oneroyalace correct me if I'm wrong!)

src/server/models/index.js
reefdog: Took the named method advice, but trying to pass in `sequelize`/`db` threw off the linter, which saw it as redefining the variables. (I tried giving the args new names, but then it choked on the effective `db[model.name] = …` bit, claiming it was [reassigning a param](https://eslint.org/docs/rules/no-param-reassign). I'm torn as to whether this is actually the case or just the linter being dumb.)

I think we can lean on lexical scoping here. It's a thing.

reefdog:
@slifty Resolved everything, though double-check my [half-measure here](https://github.com/TechAndCheck/tech-and-check-alerts/pull/42#discussion_r287036642) to see if you can accept it.

Re: seeds, short answer is seeding is a manual task (`sequelize db:seed:all`) with no environment splitting or strategy atm. I should definitely resolve that before this is merged.

src/server/migrations/20190516164909-create-claim.js
oneroyalace: `FLOAT` should be fine. Claimbuster scores  only go to three decimal places.

src/server/models/index.js
```diff
@@ -0,0 +1,37 @@
...
+const envConfig = config[env]
+
+const db = {}
+const sequelize = new Sequelize(envConfig.url, envConfig)
+
+const isFileVisible = file => file.indexOf('.') !== 0
+const isFileDifferentFromThisFile = file => file !== path.basename(__filename)
+const isFileJavaScript = file => file.slice(-3) === '.js'
+
+const registerModels = (file) => {
```
slifty: Ack sorry -- so if we do it like this the method name should be `registerModel` (since each call is registering a single model).  `registerModels` would be if you are passing the full list of tiles and the method registered each model.

I personally prefer the registerModels() approach that takes in a list because then the actual code could super clean and explicit:

```
const modelFiles = getModelFiles() // oh btw lets put the file lookup in a helper method...
registerModels(modelFiles)
registerModelAssociations()
```

reefdog: fabulous

reefdog:
Okay, made some final tweaks and cleaned up the commit to a single one. Wanna give it a last once-over @slifty?

PR #44, opened 2019-05-24
Add Jest testing framework and an example test

### LINKED ISSUE #39: Set up automated testing
We should be writing tests as we write components.  This will be much easier to do up front than it will be to retrofit automated testing.

This issue is about setting up the test framework and incorporating it into CI, not actually writing tests for any code that has been written so far (that should be made as separate issues if it applies).

* [x] Lets use ~[mocha](https://mochajs.org/) and [chai](https://www.chaijs.com/)~ [Jest](https://jestjs.io)
* [x] Lets have travis CI run the tests

reefdog:
For our testing, we’ve chosen [Jest](https://jestjs.io/). It’s fast, easy to setup, and easy to use. I’ve also created a simple example test for the `HelloWorldCrawler`.

Run tests with `yarn test`, or start a watcher with `yarn test:watcher`. Create tests for individual files as `{filename}.test.js`, and put them in a `__test__` directory alongside the file they test. E.g., `/src/server/Foo.js` is tested by `/src/server/__test__/Foo.test.js`.

- [x] Add testing framework
- [x] Add Travis CI integration

Closes #39

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,5 @@ node_js:
 
 script:
   - npm run lint
+  - npm run test
+  
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,13 @@
 {
   "presets": [
-    "@babel/preset-env"
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "node": "current"
+        }
+      }
+    ]
   ],
   "plugins": [
     "@babel/plugin-proposal-class-properties"
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,6 +7,7 @@
   "env": {
     "es6": true,
     "browser": true,
-    "node": true
+    "node": true,
+    "jest": true
   }
 }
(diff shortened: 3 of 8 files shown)


.travis.yml
```diff
@@ -5,3 +5,5 @@ node_js:
 
 script:
   - npm run lint
+  - npm run test
```
reefdog: (Actually I should probably make a special `test-ci` that's not verbose and doesn't test for coverage.)

slifty: Oh I approved and you already merged, but actually I just realized you should probably make a special `test-ci` that's not verbose and doesn't test for coverage.

reefdog: np, I'll forcefully reset master and we can recreate all of this (at a bar)

slifty: It'll be great to finally meet you.

reefdog: It'll definitely live up to expectations!

PR #52, opened 2019-06-02
Add mailer and tests

### LINKED ISSUE #22: Set up email integration
- [x] Evaluate email delivery services ([Mailgun](https://www.mailgun.com/), [Amazon SES](https://aws.amazon.com/ses/), ?)
- [x] Write integration with service
- [x] Write single-use demonstration of email delivery
- [x] Write test coverage

reefdog:
This creates a generic `Mailer` class that interacts with the Mailgun API using the `mailgun-js` npm package. It also:

* Provides test coverage for `Mailer` and its utilities
* Provides a sample send-email script, called via `yarn mailer:send-test` with an email address provided via the `-r` parameter, e.g. `yarn mailer:send-test -r test@test.com`

Closes #22

### DIFF
--- a/src/scripts/sendTestEmail.js
+++ b/src/scripts/sendTestEmail.js
@@ -19,11 +19,11 @@ if (!recipient) {
 
 const sendTestEmail = () => new Promise(() => {
   const mailer = new Mailer()
-  return Promise.resolve(mailer.send({
+  return mailer.send({
     recipient,
     subject: 'This is a test email.',
     body: 'This is the body of the email.',
-  }))
+  })
 })
 sendTestEmail().then(() => {
   process.exit()
--- a/src/scripts/sendTestEmail.js
+++ b/src/scripts/sendTestEmail.js
@@ -1,9 +1,11 @@
 import Mailer from '../server/workers/mailer'
+import { isValidEmailAddressFormat } from '../server/utils/mailer'
 
 const getRecipient = () => {
   const recipientFlag = process.argv.indexOf('-r')
   if (recipientFlag !== -1) {
-    return process.argv[recipientFlag + 1]
+    const recipient = process.argv[recipientFlag + 1]
+    if (isValidEmailAddressFormat(recipient)) return recipient
   }
   return null
 }
--- a/src/scripts/sendTestEmail.js
+++ b/src/scripts/sendTestEmail.js
@@ -1,0 +1,30 @@
+import Mailer from '../server/workers/mailer'
+
+const getRecipient = () => {
+  const recipientFlag = process.argv.indexOf('-r')
+  if (recipientFlag !== -1) {
+    return process.argv[recipientFlag + 1]
+  }
+  return null
+}
+
+const recipient = getRecipient()
+if (!recipient) {
+  console.error('Pass in a valid recipient email address with the -r flag, e.g. `-r test@test.com`')
+  console.error('(The email address must be verified with Mailgun while we are using testing subdomains.)')
+  process.exit()
+}
+
+const sendTestEmail = () => new Promise(() => {
+  const mailer = new Mailer()
+  return Promise.resolve(mailer.send({
+    recipient,
+    subject: 'This is a test email.',
+    body: 'This is the body of the email.',
+  }))
+})
+sendTestEmail().then(() => {
+  process.exit()
+}).catch((error) => {
+  console.error(error)
+})
--- a/src/scripts/sendTestEmail.js
+++ b/src/scripts/sendTestEmail.js
@@ -1,5 +1,6 @@
 import Mailer from '../server/workers/mailer'
 import { isValidEmailAddressFormat } from '../server/utils/mailer'
+import logger from '../server/utils/logger'
 
 const getRecipient = () => {
   const recipientFlag = process.argv.indexOf('-r')
@@ -12,8 +13,7 @@ const getRecipient = () => {
 
 const recipient = getRecipient()
 if (!recipient) {
-  console.error('Pass in a valid recipient email address with the -r flag, e.g. `-r test@test.com`')
-  console.error('(The email address must be verified with Mailgun while we are using testing subdomains.)')
+  logger.error('Pass in a valid and Mailgun-verified email address with the `-r` parameter.')
   process.exit()
 }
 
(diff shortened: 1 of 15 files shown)


src/scripts/sendTestEmail.js
```diff
@@ -0,0 +1,30 @@
...
+const recipient = getRecipient()
+if (!recipient) {
+  console.error('Pass in a valid recipient email address with the -r flag, e.g. `-r test@test.com`')
+  console.error('(The email address must be verified with Mailgun while we are using testing subdomains.)')
+  process.exit()
+}
+
+const sendTestEmail = () => new Promise(() => {
+  const mailer = new Mailer()
+  return Promise.resolve(mailer.send({
```
slifty: Does mailer.send() return a promise?  If so then could we just return it directly rather than wrapping in a promise?

src/server/utils/mailer.js
```diff
@@ -0,0 +1,30 @@
...
+  const hasValidBody = body => !!body && typeof body === 'string'
+
+  return (
+    hasValidRecipient(mail.recipient)
+    && hasValidSubject(mail.subject)
+    && hasValidBody(mail.body)
+  )
+}
+
+// The `mailgun-js` package lets you configure it to run in "test mode", which simulates but does
```
slifty: I don't know if this is a real preference but should multi line comments like this be /**/ -- maybe worth even doing the full blown function spec explaining the parameters and return value in more detail.

src/server/utils/mailer.js
```diff
@@ -0,0 +1,30 @@
...
+// not complete Mailgun API interactions. When this mode is enabled, however, mailgun-js runs a
+// noisy logger that logs a bunch of stuff to the console. Since we enable test mode when NODE_ENV
+// reports we're testing, this pollutes our test results with a bunch of unnecessary logging. So,
+// we create a logger that returns a bunch of useful data but doesn't actually log to console.
+export const testModeLogger = (httpOptions, payload, form) => {
```
slifty: Note: once we create a real logging system I wonder if we'll move this (or replace this).  Makes sense to put here for now.

src/server/workers/mailer/Mailer.js
```diff
@@ -0,0 +1,38 @@
...
+
+import config from '../../config'
+import { isMessageSendable, testModeLogger } from '../../utils/mailer'
+
+class Mailer {
+  constructor() {
+    this.mailgun = Mailgun({
+      apiKey: config.MAILER_API_KEY,
+      domain: config.MAILER_API_DOMAIN,
+      testMode: config.NODE_ENV === 'test',
```
slifty: 'test' should be in a constant somewhere -- in fact, we should consider creating an 'isTestEnvironment()' util method somewhere (I don't like that name, but you get the idea).  Maybe in utils/general.js

src/server/workers/mailer/Mailer.js
```diff
@@ -0,0 +1,38 @@
...
+    const mailgunData = {
+      from: config.MAILER_FROM_ADDRESS,
+      to: message.recipient,
+      subject: message.subject,
+      text: message.body,
+    }
+
+    return mailgun.messages().send(mailgunData, (error, body) => {
+      // TODO: Log (#34) or notify (#35) if things go good or bad.
+      if (error) console.log(`Error: ${error}`)
```
slifty: Omg our linter allows this?  I thought we always had to have {}

There are definitely instances I would have preferred this kind of concise one liner!  Thank you!

src/server/workers/mailer/__test__/mailerTestData.js
```diff
@@ -0,0 +1,25 @@
+export default {
```
slifty: Thank you for taking the lead on defining our test complexity -- this will be a helpful pattern in future.

src/scripts/sendTestEmail.js
```diff
@@ -0,0 +1,30 @@
+import Mailer from '../server/workers/mailer'
+
+const getRecipient = () => {
+  const recipientFlag = process.argv.indexOf('-r')
+  if (recipientFlag !== -1) {
+    return process.argv[recipientFlag + 1]
```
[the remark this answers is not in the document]
reefdog: It won't crash, it'll just return `undefined`, which gets caught by `if (!recipient)` later on. Still, the way the code reads clearly implies we'll either return a good value or a `null`, so yeah, I'll add the length condition too.

src/server/workers/mailer/Mailer.js
reefdog: Yes! I didn't know either until Smith pointed out that Prettier even seems to prefer these one-liners when possible. :tada:

src/server/utils/mailer.js
reefdog: Yeah, for sure once we have a real logger we'll want to use that. Right now the important thing is just to override the default logger, which `console.log`s. (The fact that this returns anything it all is actually pretty pointless. This could really just be a `noop`.)

I'll add a TODO about real logging!

reefdog: Oh wait we do have a real logger now. Yay Winston!

src/scripts/sendTestEmail.js
reefdog: 🤦‍♂ Yes, this works. Bless you. I was wrestling with this _so hard_ and trying so many iterations, and somehow landed on this without trying the unwrapped version.

src/scripts/sendTestEmail.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+  return null
+}
+
+const recipient = getRecipient()
+if (!recipient) {
+  console.error('Pass in a valid recipient email address with the -r flag, e.g. `-r test@test.com`')
+  console.error('(The email address must be verified with Mailgun while we are using testing subdomains.)')
```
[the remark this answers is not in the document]
reefdog: The mailer does catch it, when it checks all fields (recipient, subject, and body) here: https://github.com/TechAndCheck/tech-and-check-alerts/pull/52/commits/414021257129662501872621d358bd52a07dfef0#diff-e5b3c474b3da42780669d2858ef8f57dR20

This is just an additional check so we can provide specific guidance for calling this script.

reefdog:
Sorry, I angered Travis twice, first by not reflecting the `MAILER_*` → `MAILGUN_` env var changes, then by missing a lint warning that it noticed. Both have been fixed.

Issue #24, opened 2019-05-02
Port CNN scraping


reefdog:

slifty:
The CNN scraping has been ported!  I'm sure there will be fixes and improvements, but this is ready to be closed.

PR #54, opened 2019-06-06
Add CNN transcript list crawler

slifty:
This creates the crawler and parallel queue for the CNN transcript list pages.  These pages are basically just a list of links to actual transcripts.

This breaks out the helper functions for CNN transcripts and url parsing into its own utils/cnn.js

Related to issue #24

### DIFF
--- a/src/server/utils/cnn.js
+++ b/src/server/utils/cnn.js
@@ -1,0 +1,12 @@
+
+export const isTranscriptListUrl = url => url.startsWith('/TRANSCRIPTS/')
+  && url.endsWith('.html')
+
+export const isTranscriptUrl = url => url.startsWith('/TRANSCRIPTS/')
+  && url.endsWith('.html')
+
+export const getFullCnnUrl = (url) => {
+  if (url.startsWith('http')) return url
+  if (url.startsWith('/')) return `http://cnn.com${url}`
+  return `http://cnn.com/${url}`
+}
--- a/src/server/workers/crawlers/CnnCrawlers/index.js
+++ b/src/server/workers/crawlers/CnnCrawlers/index.js
@@ -1,4 +1,2 @@
-/* eslint-disable import/prefer-default-export */
-// We will have more exports than this, so we don't want to use export default
-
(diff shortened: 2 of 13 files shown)


src/server/utils/cnn.js
```diff
@@ -0,0 +1,12 @@
+
+export const isTranscriptListUrl = url => url.startsWith('/TRANSCRIPTS/')
+  && url.endsWith('.html')
+
+export const isTranscriptUrl = url => url.startsWith('/TRANSCRIPTS/')
+  && url.endsWith('.html')
```
reefdog: I don't see a difference between `isTranscriptUrl()` and `isTranscriptListUrl()`

slifty: There isn't one now, but that's due to laziness / circumstance, less due to the fact that there is no functional difference.

src/server/workers/crawlers/CnnCrawlers/index.js
```diff
@@ -1,4 +1,2 @@
-/* eslint-disable import/prefer-default-export */
-// We will have more exports than this, so we don't want to use export default
-
 export { default as CnnTranscriptPortalCrawler } from './CnnTranscriptPortalCrawler'
+export { default as CnnTranscriptListCrawler } from './CnnTranscriptListCrawler'
```
reefdog: This is starting to look weird but also flexing my knowledge of module exporting. Can we have two default exports?

src/server/utils/cnn.js
reefdog: Just saw your reply, after having edited my question with more detail. So to bump it: so there's a difference in the URLs, you just haven't baked that difference into the conditions yet?

src/server/workers/crawlers/CnnCrawlers/index.js
slifty: That is exactly correct.

This is just the way to allow us to access CnnTranscriptListCrawler by name when importing CnnCrawlers.

Issue #34, opened 2019-05-02
Set up logging


reefdog:
- [ ] Log running of the scraper, parser, and subscription blasts
- [ ] Logrotate logrotate

slifty:
Looks like [winston](https://www.npmjs.com/package/winston) is the node logging tool of choice.

PR #55, opened 2019-06-10
Add winston (logging)

slifty:
This adds winston as a dependency and sets up a logger for us to use throughout the project.  There is another PR that has to come next which replaces all instances of console.log with the logger.

Right now if you are not in production, all logs are written to disk in addition to being posted to terminal.  In production there are log files -- these need to be properly rotated to avoid disks being filled and killed.

We do not currently use any hosted log monitoring service, but winston makes it easy to set something like that up at some point in future.

Related to #34

### DIFF
--- a/src/server/utils/logger.js
+++ b/src/server/utils/logger.js
@@ -1,0 +1,47 @@
+import {
+  createLogger,
+  format,
+  transports,
+} from 'winston'
+
+import config from '../config'
+
+const logger = createLogger({
+  level: 'info',
+  format: format.combine(
+    format.timestamp({
+      format: 'YYYY-MM-DD HH:mm:ss',
+    }),
+    format.errors({
+      stack: true,
+    }),
+    format.splat(),
+    format.json(),
+  ),
+  defaultMeta: { service: 'tech-and-check-alerts' },
+  transports: [
+    new transports.File({
+      filename: 'logs/error.log',
+      level: 'error',
+    }),
+    new transports.File({
+      filename: 'logs/combined.log',
+    }),
+  ],
+})
+
+if (config.NODE_ENV !== 'production') {
+  logger.add(new transports.Console({
+    format: format.combine(
+      format.timestamp({
+        format: 'HH:mm:ss',
+      }),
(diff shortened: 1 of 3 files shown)


src/server/utils/logger.js
```diff
@@ -0,0 +1,45 @@
...
+      filename: 'logs/error.log',
+      level: 'error',
+    }),
+    new transports.File({
+      filename: 'logs/combined.log',
+    }),
+  ],
+})
+
+if (process.env.NODE_ENV !== 'production') {
```
reefdog: Should we hook into our `config` instead of `process.env`?

src/server/utils/logger.js
```diff
@@ -0,0 +1,45 @@
...
+    format.errors({
+      stack: true,
+    }),
+    format.splat(),
+    format.json(),
+  ),
+  defaultMeta: { service: 'tech-and-check-alerts' },
+  transports: [
+    new transports.File({
+      filename: 'logs/error.log',
```
reefdog: Is this relative to project root? Does it just assume we'll setup a symlinked directory or something?

slifty: It seems to be project root (e.g. it creates a logs directory at the root level of the project).

I think that is fine for now, though eventually we will need to be more thoughtful / have it point somewhere else in production.

src/server/utils/logger.js
slifty: OH RIGHT!

slifty: OK.

We talked a long long time (too long, in the grand scheme of things) about whether we should pull NODE_ENV from process.env or from config.

Ultimately we landed on using `process.env.NODE_ENV` -- the thinking here is that NODE_ENV is an environment variable, not a config variable.  Yes, it is set based on the environment but even if / when we move to a more effective configuration setup (e.g. yaml) for the values living in `.env.template` NODE_ENV will likely not be one of them (and will always live in `process.env.NODE_ENV`.

Maybe we're wrong. We can change it later if we need to.

Point is, not going to edit this line.

PR #58, opened 2019-06-13
Implement CNN Transcript Scraper

slifty:
Scrapers are slightly different from crawlers, though we may find reason to abstract common elements between the two since they both work with HTTP requests.

This implements our first scraper, the CNNTranscriptStatementScraper which takes a transcript URL and spits out statements (text attributed to a speaker).

These statements eventually need to be processed and passed to ClaimBuster, which will refine the statements into claims.

This does NOT create a queue for it.

Issue #24

### DIFF
--- a/src/server/utils/__test__/cnn.test.js
+++ b/src/server/utils/__test__/cnn.test.js
@@ -2,43 +2,796 @@ import {
   isTranscriptListUrl,
   isTranscriptUrl,
   getFullCnnUrl,
+  removeTimestamps,
+  removeSpeakerReminders,
+  removeDescriptors,
+  addBreaksOnSpeakerChange,
+  splitTranscriptIntoChunks,
+  getAttributionFromChunk,
+  getStatementFromChunk,
+  getNameFromAttribution,
+  getAffiliationFromAttribution,
+  extractStatementFromChunk,
+  extractStatementsFromChunks,
+  getSpeakersFromStatements,
+  cleanSpeakerName,
+  cleanStatementSpeakerNames,
+  hasIdenticalName,
+  improvesName,
+  improvesAffiliation,
+  getBestAffiliation,
+  getBestName,
+  getNormalizedSpeaker,
+  normalizeStatementSpeakers,
+  removeNetworkAffiliatedStatements,
+  removeUnattributableStatements,
 } from '../cnn'
 
-describe('isTranscriptListUrl', () => {
-  it('Should not improperly identify transcript list urls', () => {
-    expect(isTranscriptListUrl('http://google.com'))
-      .toBe(false)
-  })
-  it('Should identify transcript list urls', () => {
-    expect(isTranscriptListUrl('/TRANSCRIPTS/2019.06.01.html'))
-      .toBe(true)
-  })
-})
 
-describe('isTranscriptUrl', () => {
-  it('Should not improperly identify transcript urls', () => {
-    expect(isTranscriptUrl('http://google.com'))
-      .toBe(false)
+describe('utils/cnn', () => {
+  describe('isTranscriptListUrl', () => {
+    it('Should not improperly identify transcript list urls', () => {
+      expect(isTranscriptListUrl('http://google.com'))
+        .toBe(false)
+    })
+    it('Should identify transcript list urls', () => {
+      expect(isTranscriptListUrl('/TRANSCRIPTS/2019.06.01.html'))
+        .toBe(true)
+    })
   })
-  it('Should identify transcript list urls', () => {
-    expect(isTranscriptUrl('/TRANSCRIPTS/1906/01/cnr.20.html'))
-      .toBe(true)
-  })
-})
 
-describe('getFullCnnUrl', () => {
-  it('Should not modify a complete url', () => {
-    expect(getFullCnnUrl('http://google.com'))
-      .toBe('http://google.com')
-    expect(getFullCnnUrl('http://cnn.com'))
-      .toBe('http://cnn.com')
+  describe('isTranscriptUrl', () => {
+    it('Should not improperly identify transcript urls', () => {
+      expect(isTranscriptUrl('http://google.com'))
+        .toBe(false)
+    })
+    it('Should identify transcript list urls', () => {
+      expect(isTranscriptUrl('/TRANSCRIPTS/1906/01/cnr.20.html'))
+        .toBe(true)
+    })
   })
-  it('Should prepend a relative url', () => {
-    expect(getFullCnnUrl('TRANSCRIPTS'))
-      .toBe('http://cnn.com/TRANSCRIPTS')
+
+  describe('getFullCnnUrl', () => {
+    it('Should not modify a complete url', () => {
+      expect(getFullCnnUrl('http://google.com'))
+        .toBe('http://google.com')
+      expect(getFullCnnUrl('http://cnn.com'))
+        .toBe('http://cnn.com')
+    })
+    it('Should prepend a relative url', () => {
+      expect(getFullCnnUrl('TRANSCRIPTS'))
+        .toBe('http://cnn.com/TRANSCRIPTS')
+    })
+    it('Should prepend a relative url starting with /', () => {
+      expect(getFullCnnUrl('/TRANSCRIPTS'))
+        .toBe('http://cnn.com/TRANSCRIPTS')
+    })
   })
-  it('Should prepend a relative url starting with /', () => {
-    expect(getFullCnnUrl('/TRANSCRIPTS'))
-      .toBe('http://cnn.com/TRANSCRIPTS')
+
+  describe('removeTimestamps', () => {
+    it('Should remove timestamps of format [dd:dd:dd]', () => {
+      expect(removeTimestamps('[00:00:00] This is a test.'))
+        .toBe('This is a test.')
+      expect(removeTimestamps('This [00:00:00] is a test.'))
+        .toBe('This is a test.')
+      expect(removeTimestamps('Again, [00:00:00] this is a test.'))
+        .toBe('Again, this is a test.')
+      expect(removeTimestamps('This is a test [00:00:00].'))
+        .toBe('This is a test.')
+      expect(removeTimestamps('This is a test.[00:00:00]'))
+        .toBe('This is a test.')
+    })
+    it('Should not remove references to time', () => {
+      expect(removeTimestamps('I jumped around at 10:00:00 PM'))
+        .toBe('I jumped around at 10:00:00 PM')
+    })
+  })
+
+  describe('removeSpeakerReminders', () => {
+    it('Should remove speaker reminder inserts', () => {
+      expect(removeSpeakerReminders('I just -- JOAN DOE: -- love potatoes so much'))
+        .toBe('I just love potatoes so much')
+      expect(removeSpeakerReminders('-- JIMMY BLIMMY -- All reports indicate that I am literally on fire'))
+        .toBe('All reports indicate that I am literally on fire')
+      expect(removeSpeakerReminders('Nobody can eat that many eggs. -- KING JOHN --'))
+        .toBe('Nobody can eat that many eggs.')
+    })
+    it('Should not remove other interjections', () => {
+      expect(removeSpeakerReminders('I fear that they -- the clown army -- are going to destroy us all.'))
+        .toBe('I fear that they -- the clown army -- are going to destroy us all.')
+    })
+  })
+
+  describe('removeDescriptors', () => {
+    it('Should remove generic descriptors', () => {
+      expect(removeDescriptors('(TALKING WITH MOUTH FULL) these pickles are delicious'))
+        .toBe('these pickles are delicious')
+      expect(removeDescriptors('Here, (THROWS ARTIFICIAL LIMB) take my hand!'))
+        .toBe('Here, take my hand!')
+      expect(removeDescriptors('Follow me! (JUMPS INTO BRICK WALL)'))
+        .toBe('Follow me!')
+    })
+    it('Should remove voice over indicators', () => {
+      expect(removeDescriptors('(voice-over) behold the largest swarm of bees in the world'))
+        .toBe('behold the largest swarm of bees in the world')
+      expect(removeDescriptors('(voice- over) they turn humans into honey.'))
+        .toBe('they turn humans into honey.')
+    })
+    it('Should remove telephone indicators', () => {
+      expect(removeDescriptors('(via telephone) Hello? who is this? Why do you keep calling me?'))
+        .toBe('Hello? who is this? Why do you keep calling me?')
+      expect(removeDescriptors('Buzz buzz (via telephone)'))
+        .toBe('Buzz buzz')
+    })
+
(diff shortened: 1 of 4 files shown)


src/server/utils/cnn.js
```diff
@@ -10,3 +21,245 @@ export const getFullCnnUrl = (url) => {
...
+ */
+export const removeTimestamps = transcript => transcript
+  .replace(/\s?\[\d\d:\d\d:\d\d\]/g, '')
+  .trim()
+
+/**
+ * CNN transcripts have randomly inserted reminders of who is talking.
+ *
+ * This will remove those.
+ */
```
reefdog: Looks like a stray redundant function comment block. (L42-48 are the actual spec.)

src/server/utils/cnn.js
```diff
@@ -10,3 +21,239 @@ export const getFullCnnUrl = (url) => {
...
+ * @param  {String[]} chunks The list of transcript chunks
+ * @return {Statement[]}     The list of extracted statement objects
+ */
+export const extractStatementsFromChunks = chunks => chunks.map(
+  extractStatementFromChunk,
+)
+
+/**
+ * Returns a list of unique speakers from a list of statements
+ * @param  {Statement[]} statements The list of statement objects we want to process
```
reefdog: Need a newline between 156 and 157.

src/server/utils/cnn.js
```diff
@@ -10,3 +21,239 @@ export const getFullCnnUrl = (url) => {
...
+export const improvesName = (a, b) => !hasIdenticalName(a, b)
+  && a.name.endsWith(` ${b.name}`)
+
+export const improvesAffiliation = (a, b) => a.affiliation.length !== 0
+  && b.affiliation.length === 0
+
+export const getBestAffiliation = (a, b) => (
+  (improvesAffiliation(a, b)
+  && (hasIdenticalName(a, b)
+  || improvesName(a, b))) ? a.affiliation : b.affiliation
```
reefdog: It's not very clear to me why the name conditions would lead to improving the affiliation. Maybe that can be documented?

src/server/utils/cnn.js
```diff
@@ -10,3 +21,239 @@ export const getFullCnnUrl = (url) => {
...
+  .reduce((best, next) => {
+    const newBest = Object.assign({}, best)
+    if (improvesName(next, best)) {
+      newBest.name = getBestName(next, speaker)
+    }
+    if (improvesName(next, best)
+    || hasIdenticalName(next, best)) {
+      newBest.affiliation = getBestAffiliation(next, speaker)
+    }
+    return newBest
```
reefdog: 📎 It looks like you're trying to shortcut running `getBestName()/getBestAffiliation()` with the conditionals, yet those same conditionals are run within those functions, so we're always going to be doing twice the work. You could simplify this _and_ decrease amount of code executed with:
```
.reduce((best, next) => {
  return {
    ...best,
    name: getBestName(next, speaker),
    affiliation: getBestAffiliation(next, speaker),
  }
}, speaker)
```
Right?

src/server/utils/cnn.js
```diff
@@ -10,3 +21,239 @@ export const getFullCnnUrl = (url) => {
...
+  }, speaker)
+
+/**
+ * Normalizes all speakers in a set of statements so that the same
+ * speaker maintains the same name and attribution even if the original
+ * original transcript modified the names over time.
+ *
+ * @param  {Statement[]} statements The list of statements to normalize
```
reefdog: `Statement[]` → `Statements[]`

src/server/utils/cnn.js
```diff
@@ -10,3 +21,239 @@ export const getFullCnnUrl = (url) => {
...
+ * @return {Statements[]}           The list of normalized statements
+ */
+export const normalizeStatementSpeakers = (statements) => {
+  const speakers = getSpeakersFromStatements(statements)
+  return statements.map(statement => Object.assign({}, statement, {
+    speaker: getNormalizedSpeaker(statement.speaker, speakers),
+  }))
+}
+
+export const removeNetworkStatements = statements => statements.filter(
```
reefdog: This is unclearly named, though I figured it out by reading the code itself, so I think that's a net win? `removeStatementsByNetworkAffiliatedSpeakers()` is… verbose, but clear. Leave this one totally up to you.

src/server/utils/cnn.js
```diff
@@ -10,3 +21,239 @@ export const getFullCnnUrl = (url) => {
...
+  (statement) => {
+    const newStatement = Object.assign({}, statement)
+    newStatement.speaker.name = cleanSpeakerName(statement.speaker.name)
+    return newStatement
+  },
+)
+
+export const hasIdenticalName = (a, b) => a.name === b.name
+
+export const improvesName = (a, b) => !hasIdenticalName(a, b)
```
reefdog: I don't think you need to bother documenting the truly simplistic ones (like `hasIdenticalName()`), but I do think these qualitative judgements should explain themselves. Under what conditions does a name or affiliation generally need improving, and how does this do so?

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Process the result of an HTTP request and identify the urls
+   * to pass into the scraper pipeline.
+   *
+   * Each scraper that extends AbstractStatementScraper needs to implement its own
+   * scrapeHandler method.
+   *
+   * @param string htmlString The HTML or JSON that came from the HTTP request
+   * @return [Object] the list of statements that were scraped
```
reefdog: I'm not super versed in proper function documenting syntax, but it looks like these should be `@param {String}` and `@return {Object}`, right?

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -0,0 +1,39 @@
...
+   */
+  // (this is an abstract method and we need to define its footprint.)
+  // eslint-disable-next-line no-unused-vars
+  scrapeHandler = (responseString) => {
+    throw new Error('You implemented a statement scraper but forgot to define the scrapeHandler.')
+  }
+
+  /**
+   * Runs the crawler
+   * @return Promise a promise that returns the list of URLs that came fromthe crawl handler
```
reefdog: * Newline between L27 and 28
* "fromthe" → "from the"
* Is "Promise a promise…" a concise way of saying it's a promise wrapping a promise? If so I think it should be a little more explicit: "Returns a promise that returns a promise that returns…"

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -0,0 +1,56 @@
...
+  normalizeStatementSpeakers,
+} from '../../utils/cnn'
+
+import AbstractStatementScraper from './AbstractStatementScraper'
+
+const $ = cheerio
+
+class CnnTranscriptStatementScraper extends AbstractStatementScraper {
+  getTranscriptText = (html) => {
+    const bodyTextElements = $(html).find('.cnnBodyText')
```
reefdog: Take or leave this pattern, but I really like prefixing jQuery-wrapped (here, cheerio-wrapped) elements with `$` to make it super clear what they are and what we can do with them. Thus, this would be `const $bodyTextElements`.

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -0,0 +1,56 @@
...
+    const bodyTexts = bodyTextElements.map((i, element) => $(element).text())
+
+    if (bodyTexts.length < 2) {
+      throw new Error('The CnnTranscriptStatementScraper recieved an unexpected transcript format.')
```
reefdog: `recieved` → `received`

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -0,0 +1,56 @@
...
+import AbstractStatementScraper from './AbstractStatementScraper'
+
+const $ = cheerio
+
+class CnnTranscriptStatementScraper extends AbstractStatementScraper {
+  getTranscriptText = (html) => {
+    const bodyTextElements = $(html).find('.cnnBodyText')
+    const bodyTexts = bodyTextElements.map((i, element) => $(element).text())
+
+    if (bodyTexts.length < 2) {
```
reefdog: If we're looking for `bodyTexts[2]`, shouldn't this check be for `bodyTexts.length < 3`?

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -0,0 +1,56 @@
...
+      removeSpeakerReminders,
+      removeDescriptors,
+      addBreaksOnSpeakerChange,
+      splitTranscriptIntoChunks,
+      extractStatementsFromChunks,
+      removeNetworkStatements,
+      removeUnattributableStatements,
+      cleanStatementSpeakerNames,
+      normalizeStatementSpeakers,
+    ]
```
reefdog: If it's critical that these functions be run in this order, we should probably note that here (so future selves don't shuffle it for whatever reason, and/or know where it's safe to insert new steps).

src/server/utils/__test__/cnn.test.js
```diff
@@ -2,6 +2,29 @@ import {
...
+  improvesAffiliation,
+  getBestAffiliation,
+  getBestName,
+  getNormalizedSpeaker,
+  normalizeStatementSpeakers,
+  removeNetworkStatements,
+  removeUnattributableStatements,
 } from '../cnn'
 
 describe('isTranscriptListUrl', () => {
```
reefdog: (Pointing to a line that didn't change, I know, but here's where it matters.)

There are two things I'm doing in my test files that I'm not sure _why_ I'm doing it, I just picked it up.
* Nesting all the tests within a file in an outer describe, [like so](https://github.com/TechAndCheck/tech-and-check-alerts/blob/600906deaf16f5d684440e1dfca36a0b5bc04d64/src/server/utils/__test__/mailer.test.js#L8). So here that would be `describe('utils/cnn')` around everything. This gives an explicit and concise description to the test output, although with our focused test files this can be easily determined by the test's own filename/location.
* [Prefixing function names with `#`](https://github.com/TechAndCheck/tech-and-check-alerts/blob/600906deaf16f5d684440e1dfca36a0b5bc04d64/src/server/utils/__test__/mailer.test.js#L9). I think this might supposed to only be used when testing class methods, and I'm improperly using them everywhere?

Your thoughts on both would be 👍.

src/server/utils/cnn.js
reefdog: Actually… I said this because of the next line (`{Statements[]}`) but I think this one is actually correct and the next one needs changing, from `Statements[]` → `Statement[]`.

src/server/utils/cnn.js
```diff
@@ -10,3 +21,239 @@ export const getFullCnnUrl = (url) => {
...
+    return newBest
+  }, speaker)
+
+/**
+ * Normalizes all speakers in a set of statements so that the same
+ * speaker maintains the same name and attribution even if the original
+ * original transcript modified the names over time.
+ *
+ * @param  {Statement[]} statements The list of statements to normalize
+ * @return {Statements[]}           The list of normalized statements
```
reefdog: `Statements[]` → `Statement[]`

src/server/utils/cnn.js
slifty: The thinking was that you don't want to update an affiliation if the names are not related.

But indeed that logic doesn't belong here since it limits the utility of the function.  The logic should actually live in the normalize method, and should dictate whether this is called to begin with.

Good catch

src/server/utils/cnn.js
slifty: Decisions decisions....
I'm thinking the logic shouldn't live in the original...

So the problem is that the logic here is doing two separate comparisons.

The first is to say "is the next better than the current best"
But then the actual assignment is to compare with the ORIGINAL speaker, not the current best.

The result is that the best name is picked but ALSO that the affiliation associated with the best name is picked.  Otherwise the *first* matching affiliation would be selected.

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
slifty: agree to disagree

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
slifty: agree to disagree here too

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
slifty: IWould renaming the variable to something like stepSequence help?  Sequence implies order.

Indeed, this is a chain of functions and so order is somewhat vital.

reefdog: `stepSequence` is good. Even more self-documenting would be `stepSequenceAndByTheWayThisSequenceMattersItsNotIncidental` but I will humbly defer.

The fact that they _are_ executed in order is obvious; what wasn't clear to me was if that order was consequential and necessary. I still think an additional comment wouldn't hurt.

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
reefdog: Fair enuogh.

src/server/utils/cnn.js
reefdog: Ohh, you're right. I missed that distinction! Do what you think is right.

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
reefdog: Frair enughuo.

src/server/utils/cnn.js
slifty: I decided to keep it, but document it.

src/server/workers/scrapers/AbstractStatementScraper.js
slifty: It should have been {Promise} to indicate that the return type is a promise

src/server/utils/cnn.js
slifty: Went in the middle to `removeNetworkAffiliatedStatements`

src/server/utils/cnn.js
slifty: I'm gonna leave as is though I hear you.

src/server/utils/__test__/cnn.test.js
slifty: I'm fine with the first; I have no insight on the second...

Do you know where that # convention was documented that you found it from?  I simply don't know enough to know one way or another.

reefdog: I think I just picked it up looking at other people's tests at some point. Probably late at night.

src/server/utils/cnn.js
reefdog: 🙃

PR #59, opened 2019-06-13
Add Hello World newsletter and queue

### LINKED ISSUE #23: Set up subscription delivery
Write Hello World email subscription that...

- [x] Reads claims from database (#14)
- [x] Composes simple email template using claims
- [x] Sends template to subscribers (via Mailgun mailing list)
- [ ] Has test coverage
- [ ] Logs this happening

reefdog:
This creates a Hello World newsletter and a queue for sending it to subscribers, and stubs out future functionality.

- [x] Create newsletter concept
- [x] Create newsletter queue
- [x] Write script for sending sample newsletter (`yarn newsletter:send-test`)
- [x] Integrate real templating (Handlebars or EJS)
- [x] Log things properly
- [x] Write tests

Closes #23


src/server/newsletters/AbstractNewsletter.js
```diff
@@ -0,0 +1,64 @@
...
+    /*
+       This next line isn't entirely legible, so let's document it clearly:
+       `Handlebars.compile` receives a string of text (Handlebars-formatted markup)
+       and returns a function. You then call that function, passing in an object
+       whose properties will be made available to the template as local variables.
+       So again: `Handlebars.compile` generates a function, which we then must call
+       in order to actually receive a compiled template. (Whew.)
+    */
+    const compileTemplateWithData = Handlebars.compile(templateSource)
+    const bodyData = await this.getBodyData()
```
slifty: getBodyData is not marked as async in the abstract definition -- do we intend it to be async?

src/server/newsletters/AbstractNewsletter.js
```diff
@@ -0,0 +1,64 @@
...
+    const templateSource = fs.readFileSync(this.getPathToTemplate(), 'utf8')
+    /*
+       This next line isn't entirely legible, so let's document it clearly:
+       `Handlebars.compile` receives a string of text (Handlebars-formatted markup)
+       and returns a function. You then call that function, passing in an object
+       whose properties will be made available to the template as local variables.
+       So again: `Handlebars.compile` generates a function, which we then must call
+       in order to actually receive a compiled template. (Whew.)
+    */
+    const compileTemplateWithData = Handlebars.compile(templateSource)
```
slifty: I think this would be more self documented / we could remove the long explanation if we create a `getHandlebarsCompiler()` (either as a util or class property method).

```
const handlebarsTemplate = this.getHandlebarsTemplate(templateSource)
const bodyData = this.getBodyData()
this.body = handlebarsTemplate(bodyData)
```

Calling the output of Handlebars.compile a `template` seems to align with the Handlebars [example documentation](https://handlebarsjs.com/) (`var template = Handlebars.compile(source);`)

src/server/newsletters/AbstractNewsletter.js
```diff
@@ -0,0 +1,64 @@
...
+    const bodyData = await this.getBodyData()
+    this.body = compileTemplateWithData(bodyData)
+  }
+
+  send = () => this.render()
+    .then(() => (new Mailer()).send(this.messageData()))
+    .catch((error) => {
+      logger.error(`Could not render the newsletter ${this.constructor.name}:`)
```
slifty: I was going to say "lets make this a warning" but using the "what should developers be asked to get out of bed to fix" metric the failure to compile / send the newsletter does feel about right.

src/server/newsletters/AbstractNewsletter.js
```diff
@@ -0,0 +1,64 @@
...
+       So again: `Handlebars.compile` generates a function, which we then must call
+       in order to actually receive a compiled template. (Whew.)
+    */
+    const compileTemplateWithData = Handlebars.compile(templateSource)
+    const bodyData = await this.getBodyData()
+    this.body = compileTemplateWithData(bodyData)
+  }
+
+  send = () => this.render()
+    .then(() => (new Mailer()).send(this.messageData()))
```
slifty: This line puts a point on my discomfort with the `messageData` a function -- it's named as though it's an attribute.  We should call it `getMessageData()` but then at that point why is there a separate `render()` step (as opposed to just having a `getBody()` method which generates the body, and calling that from `getMessageData()`)?

Can / should we replace render with `getBody()`, scrap the "body" interim attribute, replace `messageData()` with:

```

const getMessageData = async () => ({
      recipient: getAddressForMailingList(this.getMailingList()),
      subject: this.getSubject(),
      body: await this.getBody(),
})
```

src/server/constants.js
```diff
@@ -1,8 +1,19 @@
...
+
+export const MAILING_LISTS = {
+  DEV: 'DEV',
+  TEST: 'TEST',
+  PRIMARY: 'PRIMARY',
+  NC: 'NC',
+}
+
+export const MAILING_LIST_ADDRESSES = {
+  DEV: 'dev',
```
slifty: Lets call this "DEVELOPERS" and "TESTERS" so they don't read like an environment. (same above too)

src/server/newsletters/HelloWorldNewsletter.js
```diff
@@ -0,0 +1,26 @@
...
+
+  getPathToTemplate = () => `${__dirname}/templates/helloWorld.hbs`
+
+  getSubject = () => 'Hello World: The Newsletter'
+
+  // We aren't yet performing newsletter-specific scheduling.
+  // getSchedule = () => '0 9 * * *'
+
+  getBodyData = async () => {
+    const claims = await models.Claim.findAll({
```
slifty: Would  love for this to actually be its own `getClaims()` method and getBodyData then becomes:

```
getBodyData = async () => ({
  claims: await this.getClaims()
})
```

src/server/utils/mailer.js
```diff
@@ -15,6 +18,10 @@ export const isMessageSendable = (mail) => {
   )
 }
 
+export const getAddressForMailingList = (listKey) => {
+  if (!Object.prototype.hasOwnProperty.call(MAILING_LIST_ADDRESSES, listKey)) return null
```
slifty: omg eslint this makes NO SENSE WHAT IS HAPPENING.

slifty: In seriousness.... could you wrap this in a well named helper method...

e.g. `doesMailingListHaveAddress()`

src/server/utils/mailer.js
```diff
@@ -15,6 +18,10 @@ export const isMessageSendable = (mail) => {
   )
 }
 
+export const getAddressForMailingList = (listKey) => {
```
slifty: Pedantic and maybe just preference: this might be better named as `getMailingListAddress` since get{attribute} feels more natural than get{attribute}for{thing} <-- and addressForMailingList isn't what we would label the attribute.

yarn.lock
```diff
@@ -2515,6 +2515,11 @@ date-now@^0.1.4:
   resolved "https://registry.yarnpkg.com/date-now/-/date-now-0.1.4.tgz#eaf439fd4d4848ad74e5cc7dbef200672b9e345b"
   integrity sha1-6vQ5/U1ISK105cx9vvIAZyueNFs=
 
+dayjs@^1.8.14:
```
slifty: Have you considered momentJS?

src/server/workers/mailer/Mailer.js
```diff
@@ -29,10 +29,15 @@ class Mailer {
...
-      if (error) logger.error(error)
-      else if (body) logger.info(body)
-    })
+    const logMessage = `"${mailgunData.subject}" from "${mailgunData.from}" to "${mailgunData.to}"`
+
+    return mailgun.messages().send(mailgunData)
+      .then(() => {
+        logger.info(`Mailgun received ${logMessage} successfully.`)
+      }).catch((error) => {
+        logger.error(`Mailgun failed to send ${logMessage}.`)
```
slifty: This does feel more like a warning than an error.

package.json
```diff
@@ -42,7 +43,9 @@
   "dependencies": {
     "bull": "^3.9.1",
     "cheerio": "^1.0.0-rc.3",
+    "dayjs": "^1.8.14",
```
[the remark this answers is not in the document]
slifty: wait actually in seriousness I'm realizing that was added in this PR but isn't actually used anywhere.

src/server/newsletters/AbstractNewsletter.js
```diff
@@ -0,0 +1,64 @@
...
+  getBodyData = () => {
+    throw new Error('You extended AbstractNewsletter but forgot to define getBodyData()')
+  }
+
+  messageData = () => {
+    if (!this.body) {
+      throw new Error('You have to call the render() method to render newsletter content before generating message data.')
+    }
+    return {
+      recipient: getAddressForMailingList(this.getMailingList()),
```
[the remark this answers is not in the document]
slifty: Actually though: should we consider thinkingof mailingListAddress as an attribute of AbstractNewsletter?  In other words should we implement a this.getMailingListAddress() on the AbstractNewsletter() <-- one that does not take in a mailing list parameter but rather leverages this.getMailingList() internally?

This could be instead of *or* in addition to the util...

src/server/newsletters/AbstractNewsletter.js
slifty: (on second thought this really should be a warning, since it is a recoverable error and doesn't break the entire system)

src/server/newsletters/AbstractNewsletter.js
slifty: (I just read the commit message associated with this and 🍭)

package.json
```diff
@@ -21,7 +21,8 @@
     "queue:jobs:schedule": "babel-node -- src/scripts/scheduleJobs",
     "queue:jobs:unschedule": "babel-node -- src/scripts/unscheduleJobs",
     "queue:jobs:run:cnn-portal-crawler": "babel-node -- src/scripts/runCrawler",
-    "mailer:send-test": "babel-node -- src/scripts/sendTestEmail"
+    "mailer:send-test": "babel-node -- src/scripts/sendTestEmail",
+    "newsletter:send-test": "babel-node -- src/scripts/sendTestNewsletter"
```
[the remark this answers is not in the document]
reefdog: See, I was thinking the inverse (obviously): that we may have more worker-specific commands and utilities, and so it made sense to prefix by worker name. But you may be right; what I'm really doing here is performing a live test of a specific worker, which _is_ something we likely want to do commonly and repeatedly. Hmm.

If it's okay, I'll leave this alone until we have more examples to draw a pattern from.

package.json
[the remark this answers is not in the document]
reefdog: Hahah you're right! I deleted our one use of it in 0d8f9e7efab66c68ddc5c7e1eaab1ed172b83b00 but forgot to remove the library. Will delete! Also, it seems we have Moment whether we want it or not b'c Sequelize requires it.

src/server/newsletters/AbstractNewsletter.js
```diff
@@ -0,0 +1,64 @@
...
+
+  getPathToTemplate = () => {
+    throw new Error('You extended AbstractNewsletter but forgot to define getPathToTemplate()')
+  }
+
+  getSubject = () => {
+    throw new Error('You extended AbstractNewsletter but forgot to define getSubject()')
+  }
+
+  // We aren't yet performing newsletter-specific scheduling.
```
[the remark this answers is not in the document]
reefdog: I'll remove it; it's easy to recreate and there's no deep problem-solving we need to keep around for reference.

package.json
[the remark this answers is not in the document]
slifty: FINE.

(your point here makes sense and is legit too -- I was just navel gazing anyway; definitely perfectly fine for now regardless of potential futures)

src/server/newsletters/AbstractNewsletter.js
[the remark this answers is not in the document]
reefdog: That makes total sense and I like it!

src/server/newsletters/AbstractNewsletter.js
reefdog: Good question. In theory, we could compose the body data for a specific newsletter synchronously. In reality, all our real newsletters will query the database for their body data, which needs to happen asynchronously. So should the abstract definition match theory or all known realities? 🤔

src/server/newsletters/AbstractNewsletter.js
reefdog: That's an excellent solution for something I hit my head against the wall about for a while. :pray:

src/server/newsletters/AbstractNewsletter.js
```diff
@@ -0,0 +1,64 @@
...
+  //   throw new Error('You extended AbstractNewsletter but forgot to define getSchedule()')
+  // }
+
+  getBodyData = () => {
+    throw new Error('You extended AbstractNewsletter but forgot to define getBodyData()')
+  }
+
+  messageData = () => {
+    if (!this.body) {
+      throw new Error('You have to call the render() method to render newsletter content before generating message data.')
```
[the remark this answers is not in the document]
reefdog: Is poor design a reason? Because the reason is poor design. 😬 (You've suggested improvements elsewhere that will resolve this, so implementing those changes.)

src/server/newsletters/AbstractNewsletter.js
reefdog: As you originally noted, errors while sending a newsletter _should_ throw an error, because we Really Want To Know about them. But there are other places, like generic mailer failures, that should be warnings. I'm going to make a log level pass to fix all this. 👍

src/server/newsletters/HelloWorldNewsletter.js
reefdog: Good call.

src/server/utils/mailer.js
reefdog: Good point. And honestly we should move the crazy Object.property thing to a generic `hasKey` utility, too.

src/server/newsletters/AbstractNewsletter.js
slifty: Short Story: The abstract method should be async.

Long Story:  the point of abstract methods is for other objects to be able to predict the behavior of implemented methods from an API level (async is an API level decorator).  In a strongly typed language that supports abstraction you would not be allowed to override a method while also changing the API of that method.

reefdog: Of cooouuurse. That's very sensible. Thanks!

reefdog:
Ready for re-review, @slifty!

src/server/newsletters/AbstractNewsletter.js
```diff
@@ -0,0 +1,98 @@
...
+   *
+   * OVERRIDE WHEN EXTENDING
+   *
+   * @return {object} The object that will be passed to the templating function
+   */
+  getBodyData = async () => {
+    throw new Error('You extended AbstractNewsletter but forgot to define getBodyData()')
+  }
+
+  getMailingListAddress = () => {
```
slifty: Even though this isn't abstract we *should* document it.

reefdog:
Oh man. I totally meant to squash all these many many commits. 🤦‍♂ I'll do that in the future for such huge PRs.

PR #62, opened 2019-06-14
Add cnn transcript scraper queue

slifty:
This creates a queue for the CNN transcript statement scraper.  It also includes a few little fixes to the scraper that came up in the first 30 seconds of testing.

Oh, it also has a little tweak to our logger so that errors render the stack trace; that should have always been there.

Related to Issue #24

### DIFF
--- a/src/server/utils/cnn.js
+++ b/src/server/utils/cnn.js
@@ -64,7 +64,7 @@ export const removeDescriptors = transcript => transcript
  * @return {String}            The modified transcript with line breaks inserted.
  */
 export const addBreaksOnSpeakerChange = transcript => transcript
-  .replace(/([.?!\-)\]])\s*([.,A-Z\s]*:)/g, '$1\n$2')
+  .replace(/([.?!\-)\]])\s*([.,A-Z\s"()]*:)/g, '$1\n$2')
 
 
 /**
@@ -77,7 +77,7 @@ export const addBreaksOnSpeakerChange = transcript => transcript
 export const splitTranscriptIntoChunks = transcript => transcript.split('\n')
 
 // Used to identify the speaker section of a chunk
-const chunkSpeakerRegex = /[.,A-Z\s]+[:]/
+const chunkAttributionRegex = /[.,A-Z\s"()]+[:]/
 
 /**
  * Extract the speaker name / affiliation from a chunk.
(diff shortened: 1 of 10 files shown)


src/server/utils/cnn.js
```diff
@@ -64,7 +64,7 @@ export const removeDescriptors = transcript => transcript
  * @return {String}            The modified transcript with line breaks inserted.
  */
 export const addBreaksOnSpeakerChange = transcript => transcript
-  .replace(/([.?!\-)\]])\s*([.,A-Z\s]*:)/g, '$1\n$2')
+  .replace(/([.?!\-)\]])\s*([.,A-Z\s"()]*:)/g, '$1\n$2')
```
reefdog: Just a question: attributions are always uppercase, right? That's why you can get away with this (and in fact one way we recognize attributions)?

slifty: Correct!

src/server/utils/cnn.js
```diff
@@ -77,17 +77,19 @@ export const addBreaksOnSpeakerChange = transcript => transcript
 export const splitTranscriptIntoChunks = transcript => transcript.split('\n')
 
 // Used to identify the speaker section of a chunk
-const chunkSpeakerRegex = /[.,A-Z\s]+[:]/
+const chunkAttributionRegex = /[.,A-Z\s"()]+[:]/
```
reefdog: This might start an unfortunate chain reaction of work, but having to change a regex pattern (here, `[.,A-Z\s"()]`) for the same thing in two places likely means we should capture it in a constant.

However, composing a regex pattern using a variable means switching to the `new Regex()` syntax rather than the slash shorthand (chain reaction!), so, let's open an issue and move on unless you think it's a Real Good Idea to tackle it now.

src/server/utils/cnn.js
```diff
@@ -115,7 +118,7 @@ export const getNameFromAttribution = attribution => attribution
  *                                  the person's affiliation
  */
 export const getAffiliationFromAttribution = attribution => (
-  (attribution.match(/,([A-Z\s,]*)/)
+  (attribution.match(/,([A-Z\s,"()]*)/)
```
reefdog: Another place we would rely on the pattern constant.

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -1,6 +1,8 @@
 import cheerio from 'cheerio'
 
 import {
+  isTranscriptUrl,
```
reefdog: Do you know enough about what makes a transcript URL different from a transcript list URL that you could start improving that check and differentiating them?

slifty: Yes but I refuse to do so JUST BECAUSE I know it bothers you so much.

src/server/utils/cnn.js
slifty: My thought process was: "it sure would be nice to only have to change this once.  Oh well I guess that's impossible"

Issue #66, opened 2019-06-18
Add memory to our scrapers


slifty:
We need a way to prevent the scrape of the same pages twice.  This behavior won't be the same for each type of scraper (for instance we want to crawl the CNN transcript portal every day, but we never want to scrape the same transcript more than once), but the functionality should be part of the abstract classes and configured at the implementation level.

## Where should we store

We could either...

1. Use SQL to cache the list of URLs that have been scraped (ID, URL, scraperClass, dateScraped).

2. Use the file system; write to a {urlHash}.txt file upon successful scrape, and maybe even store the results of the scrape in the file.

3. Use Redis instead of a filesystem.

This feels like a non-optimal use for SQL, but since we're looking at thousands of scrapes (not millions or billions) anything could work.  I'm leaning towards Redis since we already use Redis for the scraping process and it has most benefits of the filesystem but is more engineered / can scale as needed.

## Allowing configuration

Two approaches to configuration:

1. Have the default run() behavior check the database first and reject the scrape if it is a duplicate; provide the ability to force the run by passing in run(true)

2. Create the ability to set a repetition frequency in the overriding class (similar to the way we do schedules)

reefdog:
Some thoughts:

1. Read performance is most important. We don't care how quickly it writes (within reason).
2. Not that I'm otherwise arguing for SQL over Redis, but "since we're looking at thousands of scrapes (not millions or billions) anything could work" implies SQL is more appropriate at K-scale than M/B-scale. Why is that? MySQL/PostgreSQL are… rather proven at those scales.
3. This is basically a flat denylist: we configure which types of URLs, once scraped, shouldn't be scraped again unless forced. Add those to the denylist. Upon run, filter the scrapelist by the denylist. If we want to _also_ keep a record of every URL scraped (regardless of whether it can be re-run) and when, that's fine too, but without a clear use for that data right now we should make sure it doesn't take precedence over the immediate need. (But of course, if it's a little more work for a lot of long-term historical record we want to keep, now's the best time.)

slifty:
Re #2: I could be totally off base!  My understanding is that postgres is just a bit slower than redis for this kind of simple lookup, which doesn't matter at K scale but could start to impact performance at M / B.  Some of this is because SQL has a lot more guarantees around consistency and not losing data.  We don't really care about those things for this use case because worst case we just scrape something an extra time.

[...]

PR #67, opened 2019-06-18
Add redis to the project

slifty:
This system already required redis in order to operate, since redis is what we use for queues.  This makes it possible to interact with redis in our application.

Related to issue #66 since we're going to be using redis to store crawl history

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "dotenv": "^8.0.0",
     "mailgun-js": "^0.22.0",
     "pg": "^7.10.0",
+    "redis": "^2.8.0",
     "request": "^2.88.0",
     "request-promise": "^4.2.4",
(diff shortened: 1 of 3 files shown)


src/server/utils/redis.js
```diff
@@ -0,0 +1,8 @@
+import redis from 'redis'
+import config from '../config'
+
+const client = redis.createClient({
+  url: config.REDIS_URL,
+})
+
+export default client
```
reefdog: Yeah, definitely feeling like this belongs in `src/server/redis/index.js` (or `src/server/redis/client.js`) than in `utils`.

slifty: OK Merge it is then thanks!!

reefdog: 💁🏼‍♂️

PR #70, opened 2019-06-19
Don’t do verbose coverage testing on CI

### LINKED ISSUE #69: Make a faster test-ci script
Our default [`test`](https://github.com/TechAndCheck/tech-and-check-alerts/blob/09bee8da319b62ad659feb948330710aef19ab78/package.json#L17) script includes the `--verbose` and `--coverage` options. CI doesn't need to do these, so we should make a `test-ci` version without those options and run that one [here](https://github.com/TechAndCheck/tech-and-check-alerts/blob/09bee8da319b62ad659feb948330710aef19ab78/.travis.yml#L8) instead.

This will take two seconds but I want to sigh use the proper dog-and-pony show.

reefdog:
CI doesn’t benefit from coverage testing or verbose results, so configure Travis to use a special version without those options.

Closes #69

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "lint": "./node_modules/.bin/eslint 'src/**/*.js'",
     "test": "jest --verbose --coverage",
     "test:watch": "jest --verbose --watch",
+    "test:ci": "jest",
     "migrate": "sequelize db:migrate",
     "queue:jobs:list": "babel-node -- src/scripts/listScheduledJobs",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
     "lint": "./node_modules/.bin/eslint 'src/**/*.js'",
     "test": "jest --verbose --coverage",
     "test:watch": "jest --verbose --watch",
+    "test:ci": "jest",
```
slifty: I was going to jokingly ask for a change to this line, but it would have been in jest.

Issue #46, opened 2019-05-30
Add tests for queue system


reefdog:
We gained a testing framework in #44, but #45 had already been queued up (lolol 👏) and we need it for moving forward, so it currently lacks tests. But being a critical bit of infrastructure, we should add tests for it.

PR #73, opened 2019-06-20
Add scraper memory

slifty:
This commit builds out a basic memory for the scrapers, so that our system knows whether or not a given URL has been processed by a given scraper in the past.

This introduces a new dependency (`dayjs`) for working with time

## Item 1: Remembering things
We now keep track of scrapes that resulted in error OR success, and also logs the outcome of the scrape (or the error if it failed).

Here's what's being stored:
- (base) the timestamp of the most recent scrape attempt
- (successLedger) the list of timestamps of successful scrapes
- (failureLedger) the list of timestamps of failed scrapes
- (errors) the contents of errors (on failure)
- (results) the contents of results (on success)

An example Redis key set for a url with a single successful scrape:
image

## Item 2: Using that memory
Furthermore this PR applies a very first "attempt" at logic to avoid scraping the same page twice.  As we discussed in #66 there is room for more.  I also don't think the logic should necessarily live in the queue processor but rather should probably be *calculated* at a scraper level.  In other words, we should have something like a `shouldScrape()` command at the AbstractScraper level.

In the spirit of not over engineering (for once) I would like to defer that to a future issue, and just have the basic "avoid the noise" level check here which should cover our baseline use cases. This is captured in #74 

NOTE: I also took out some tests for the scrapers because they were skeletons anyway and were breaking CI due to the reliance on redis.  Developing a real test suite for the queues and scrapers is captured in #46


package.json
```diff
@@ -42,6 +42,7 @@
   "dependencies": {
     "bull": "^3.9.1",
     "cheerio": "^1.0.0-rc.3",
+    "dayjs": "^1.8.14",
```
reefdog: Just leaving a quick note that if we ever need to inflate our package size and decrease performance we should consider Moment.js.

src/server/workers/scrapers/AbstractScraper.js
```diff
@@ -24,21 +28,122 @@ class AbstractScraper {
     throw new Error('You implemented a scraper but forgot to define the scrapeHandler.')
   }
 
+  /**
+   * Returns a list of keys that are used in redis to store information about
+   * the scrape.
+   *
+   * - base: stores the most recent scrape result
```
reefdog: Is `base` a generic/common name for a thing? I'm not familiar with it and it's not jumping out to me as legible/self-documenting.

Edit: Oh a little later on I realize what this is.

src/server/workers/scrapers/AbstractScraper.js
```diff
@@ -24,21 +28,122 @@ class AbstractScraper {
...
+   * @return {null}
+   */
+  registerScrapeSuccess = () => {
+    redisClient.lpush(
+      this.getRedisKeys().successLedger,
+      Date.now(),
+    )
+    redisClient.set(
+      this.getRedisKeys().base,
+      Date.now(),
```
reefdog: Since we would expect both of these dates to be the same, I say calculate `Date.now()` once at the top of the function and output it both places by reference. We don't want them to differ even by microseconds.

src/server/workers/scrapers/AbstractScraper.js
```diff
@@ -24,21 +28,122 @@ class AbstractScraper {
...
+   * @return {null}
+   */
+  registerScrapeError = () => {
+    redisClient.lpush(
+      this.getRedisKeys().failureLedger,
+      Date.now(),
+    )
+    redisClient.set(
+      this.getRedisKeys().base,
+      Date.now(),
```
reefdog: Same comment as above re: date drift.

src/server/workers/scrapers/AbstractScraper.js
```diff
@@ -24,21 +28,122 @@ class AbstractScraper {
...
 
+  /**
+   * Returns a list of keys that are used in redis to store information about
+   * the scrape.
+   *
+   * - base: stores the most recent scrape result
+   * - successLedger: stores a set of scrape dates that succeeded
+   * - failureLedger: stores a set of scrape dates that failed
+   * - results: stores a history of scrape results
+   * - errors: stores a history of scrape errors
```
reefdog: These names are a little funny. They're arguably all ledgers, just of different things. So I'd say append `Ledger` to them all, or differentiate based on contents: `successDates`/`failureDates`, then you could either keep `results`/`errors` as-is or make them like `resultBodies`/`errorBodies`. Thoughts?

src/server/workers/scrapers/AbstractScraper.js
```diff
@@ -24,21 +28,122 @@ class AbstractScraper {
...
+   *
+   * @return {[type]} [description]
+   */
+  getRedisKeys = () => {
+    const base = `scrape:${this.constructor.name}:${this.scrapeUrl}`
+    return {
+      base,
+      successLedger: `${base}:successLedger`,
```
reefdog: Oh now I understand `base` a little more. Hmm. Maybe `basePrefix` or `keyPrefix`?

src/server/queues/cnnTranscriptListCrawlerQueue/cnnTranscriptListCrawlerJobProcessor.js
```diff
@@ -1,11 +1,23 @@
...
-
+import CnnTranscriptStatementScraper from '../../workers/scrapers/CnnTranscriptStatementScraper'
 import cnnTranscriptStatementScraperQueueDict from '../cnnTranscriptStatementScraperQueue'
 
 const statementScraperQueue = cnnTranscriptStatementScraperQueueDict.factory.getQueue()
 
 const scrapeTranscriptUrl = url => statementScraperQueue.add({ url })
 
+const processTranscriptListUrl = async (url) => {
+  const scraper = new CnnTranscriptStatementScraper(url)
```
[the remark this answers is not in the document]
slifty: Oh this should have been processTranscriptUrl my mistake.

This is the place where we avoid adding jobs to the queue to scrape transcripts that have been scraped before.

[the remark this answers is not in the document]
reefdog: Ah! Then that makes total sense, and just needs the name change to `processTranscriptUrl()`. 👍

src/server/utils/redis.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+export const promisifyClient = (client) => {
+  const methods = [
+    'get',
+    'lrange',
+  ] // The list of redux methods we want to promisify
+
+  return methods.reduce((oldClient, method) => {
+    const newClient = oldClient
+    newClient[`${method}Async`] = promisify(oldClient[method])
```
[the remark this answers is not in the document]
slifty: So really this is just creating a bootstrap version of [Bluebird's promisifyAll](http://bluebirdjs.com/docs/api/promise.promisifyall.html) (which I think would also fail the grep test).

The only solution I can think of would be to take out the `.reduce` approach and simply say

newClient.getAsync = promisify(client.get)
newClient.lrangeAsync = promisify(client.lrange)

The smell introduced there is DRY violation but I suppose everything is a compromise.

src/server/workers/scrapers/AbstractScraper.js
slifty: Also: right now I actually use that base to store the most recent scrape attempt; I am realizing it might be more appropriate to label that explicitly.

src/server/utils/redis.js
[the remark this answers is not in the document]
reefdog: I really do. Metaprogramming is too convenient to discard, but I really do foresee being frustrated when trying to track down the function definition of one of those new methods far in future. I would even include the output name in the input object definition, thus:

```
const methods = [
    'get', // => `getAsync`
    'lrange', // => `lrangeAsync`
  ] // The list of redux methods we want to promisify
```

src/server/workers/scrapers/AbstractScraper.js
slifty: I like the content approach.

src/server/workers/scrapers/AbstractScraper.js
reefdog: Also a good point!

src/server/utils/redis.js
```diff
@@ -0,0 +1,46 @@
+import dayjs from 'dayjs'
+import { promisify } from 'util'
+
+/**
+ * Returns a version of the redis client with promise based
+ * versions of key methods.
+ * @param  {Object} client The redis client that came from the redis client factory
```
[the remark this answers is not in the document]
slifty: The first version of this that I wrote had a newline.  It's when I had to rewrite it that I forgot to add the newline..........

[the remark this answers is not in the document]
reefdog: Please make sure every one of your PRs contains one of these, just to keep me on my toes.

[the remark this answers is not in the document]
reefdog: Also a `TranscriptUrl`/`TranscriptListUrl` gotcha.

src/server/utils/__test__/redis.test.js
```diff
@@ -0,0 +1,56 @@
...
+      expect(time.format('YYYY-MM-DD'))
+        .toEqual('2019-06-20')
+    })
+    it('Should return null for invalid timestamps', () => {
+      expect(parseTime(''))
+        .toBe(null)
+      expect(parseTime('I Love Lamp'))
+        .toBe(null)
+    })
+  })
```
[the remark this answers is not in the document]
slifty: I made a test -- it does not seem to work.  Since we're just generating these with Date.now() we don't need to support different kinds of timestamps at this stage.  That said, parseTime() is probably badly named.  Really this is about extracting a `Date` generated timestamp from a Redis response which may or may not have an actual timestamp (e.g. might be empty or null).

Meh.

src/server/workers/scrapers/AbstractScraper.js
slifty: @reefdog do you think it is a problem to call it `successDates` if the content being stored is a timestamp (e.g. more specific than a date)?

Should I say 'successTimestamps'

Also I'm going to rename base to `latestAttempt` -- should that be `latestAttemptDate`

reefdog: I _do_ like being specific with `Timestamps`, but two thoughts:

1. We've sailed that boat a lot already given that, like, `Date.now()` returns a timestamp and people use "date" to mean "time" a lot. 😵
2. One convention is `At`, as in `createdAt`/`updatedAt`. But this isn't useful because `successesAt` or `successAts` are both really silly.

Both those thoughts lead me to say let's do the right thing, but can we do `Times` instead of `Timestamps`? Or is that too easily confused as "number of times"? (In which case sure, `Timestamps`.)

And yes, whatever we do here should probably be appended to `latestAttempt` as well.

src/server/workers/scrapers/AbstractScraper.js
slifty: It is still a bad name -- the prefix isn't useful anywhere, I was just hiding the latest scrape date in that key.  I'm updating to be more explicit in that label

src/server/workers/scrapers/AbstractScraper.js
slifty: Times works, but I'll just use Ats.

src/server/utils/__test__/redis.test.js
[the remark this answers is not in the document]
reefdog: Sounds good on the test; I didn't track this back to realize we were the only source of input. We'll want to shore it up a little more once we're passing in times gleaned from scrapes, but this is fine for now.

And yeah, `parseTime` isn't fantastic (not least because we aren't really parsing the time, we're just converting a string to a Day.js object with a little validation and fallback). But I say leave it for now.

PR #88, opened 2019-07-17
Adjust claims model

slifty:
This PR removes the speaker model / table and adds the following columns to the claims table:

* speaker_name
* speaker_affiliation
* source
* scraper_code

It also comes with some tech debt fixing by adding babel to sequelize and fixing a typo in a past migration.

Finally, it breaks some new ground for our migrations by creating a model for running more than one action in a migration step.  We now have a mechanism to [run promises sequentially](https://decembersoft.com/posts/promises-in-serial-with-array-reduce/) and also it uses a transaction so that if a step fails the whole migration is undone rather than having an awkward partial migration.

It also (probably completely unnecessarily) migrates the data both on the up AND down.  This took way too long to implement so YOU HAD BETTER APPRECIATE IT GOSH DARN IT...


src/server/migrations/20190703142652-add-affiliation.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     ),
 
   down: queryInterface => queryInterface
-    .reoveColumn(
+    .removeColumn(
```
reefdog: Please stop belittling me.

src/server/models/claim.js
```diff
@@ -2,13 +2,12 @@ module.exports = (sequelize, DataTypes) => {
   const Claim = sequelize.define('Claim', {
     content: DataTypes.TEXT,
     claimedAt: DataTypes.DATE,
-    sourceUrl: DataTypes.STRING(1024),
+    canonicalUrl: DataTypes.STRING(1024),
+    scraperCode: DataTypes.STRING(1024),
+    source: DataTypes.STRING(1024),
+    speaker_name: DataTypes.STRING(1024),
+    speaker_affiliation: DataTypes.STRING(1024),
```
reefdog: Can you camelCase these two?

src/server/models/claim.js
```diff
@@ -2,13 +2,12 @@ module.exports = (sequelize, DataTypes) => {
   const Claim = sequelize.define('Claim', {
     content: DataTypes.TEXT,
     claimedAt: DataTypes.DATE,
-    sourceUrl: DataTypes.STRING(1024),
+    canonicalUrl: DataTypes.STRING(1024),
+    scraperCode: DataTypes.STRING(1024),
```
reefdog: Do you note anywhere what `scraperCode` is?

src/server/utils/index.js
```diff
@@ -4,3 +4,8 @@ import { ENV_NAMES } from '../constants'
 export const isDevelopmentEnv = () => config.NODE_ENV === ENV_NAMES.DEVELOPMENT
 export const isTestEnv = () => config.NODE_ENV === ENV_NAMES.TEST
 export const isProductionEnv = () => config.NODE_ENV === ENV_NAMES.PRODUCTION
+
+export const runPromiseSequence = promiseArray => promiseArray.reduce(
```
reefdog: I'm torn on whether "sequence", **as previously discussed**, is a sufficiently self-documenting term. I fully understand that, by definition, a break in the sequence means the sequence stops. But I can also seem someone running across this and missing that, and thinking it's just sugar for executing a bunch of promises (for some reason where `Promise.all` isn't appropriate).

Anyway, I think explicitly documenting the behavior of this wouldn't hurt.

src/server/models/claim.js
slifty: No.

src/server/models/claim.js
slifty: No.

reefdog: Counterpoint:

src/server/models/claim.js
reefdog: Counterpoint: 👊🏼

src/server/utils/index.js
reefdog: Okay so ignore my complaint about "sequence", I think this is just me being obtuse.

But I do think this function could be a little less magical by slightly copying the style from [this SO answer](https://stackoverflow.com/a/30823708):

```
export const runPromiseSequence = promiseArray => promiseArray.reduce(
  (current, next) => current.then(next),
  Promise.resolve(),
)
```

This way you're starting the reduce with a junk `Promise.resolve()` rather than modifying the `promiseArray` param, which is actually an ESLint violation I think?

reefdog:
If #93 gets approved and merged before this, you'll want to make sure to tackle adapting the `PrimaryNewsletter` to use these new models. The specific points:

- Remove the [Speaker association](https://github.com/TechAndCheck/tech-and-check-alerts/blob/c120bf2d2f5b7099b0043298943b9fa2285071a3/src/server/newsletters/PrimaryNewsletter.js#L34-L37)
- Update the `speaker.name` and `sourceUrl` references in the templates ([HTML](https://github.com/TechAndCheck/tech-and-check-alerts/blob/c120bf2d2f5b7099b0043298943b9fa2285071a3/src/server/newsletters/templates/primary.hbs) and [text](https://github.com/TechAndCheck/tech-and-check-alerts/blob/c120bf2d2f5b7099b0043298943b9fa2285071a3/src/server/newsletters/templates/primaryText.hbs))
- For extra credit, add the new claim platform (CNN/Twitter) and source (for CNN only, the show name) details inside the `<a class="claim__permalink">` tag (but outside the `<cite>`); for proper styling, the platform should also be added to the `<li class="claim">` tag as the `data-platform` attribute in the HTML version, ala: `<li class="claim" data-platform="CNN">`

src/server/models/claim.js
slifty: I thought about this for a while on vacation and decided I'll give into your unreasonable request.

src/server/models/claim.js
slifty: OK so -- this is meant to indicate what scraper generated the claim (e.g. cnn, twitter, etc).  Right now there are worker queue names but I don't think that's really what we want to use here.  I think I'll end up making a ScraperCodes in a `workers/scrapers/constants.js` file that doesn't exist yet.  I'll add a short comment int he model to document this intent.

src/server/utils/index.js
slifty: yeah that's much cleaner -- thank you.

PR #93, opened 2019-07-26
Add primary newsletter

reefdog:
This PR, which is currently a draft, adds the real primary newsletter:

- Creates the `PrimaryNewsletter` class
- Adds additional templating and styling
- Ensures we don't accidentally send it to real people while testing/developing


src/server/newsletters/PrimaryNewsletter.js
```diff
@@ -0,0 +1,45 @@
...
+    order: [
+      ['claimBusterScore', 'DESC'],
+    ],
+    include: [{
+      model: Speaker,
+      as: 'speaker',
+    }],
+  }).then(tvClaims => tvClaims))
+
+  getBodyData = async () => ({
```
slifty: ahhh, I see -- we actually define getBodyData here and so we want the cache logic in the abstract method.

What if the Abstract method were getBodyData and the override method were something like `generateBodyData`, `loadBodyData`, or `createBodyData`

src/server/newsletters/templates/primary.hbs
```diff
@@ -0,0 +1,22 @@
...
+  <ul class="claims">
+  {{#each tvClaims}}
+    <li class="claim">
+      <div class="claim__metadata">
+        <a href="{{sourceUrl}}" class="claim__permalink">
+          <cite class="claim__speaker">
+            {{speaker.name}}
+          </cite></a>:
+      </div>
+      <blockquote class="claim__content">{{content}}</blockquote>
```
slifty: Since we know fastmail does funky blockquote styling do we need / want something that will clear that kind of styling out? (this would make sense as an issue for future attention)

src/server/utils/templates.js
```diff
@@ -34,29 +37,26 @@ export const stripHTMLTags = html => sanitizeHTML(html, {
...
+// `decodeAmpersands()` should be unnecessary, but sanitizeHTML's `decodeEntities` doesn't work
+const decodeAmpersands = text => text.replace(/&amp;/g, '&')
+
+/**
+ * Cleans HTML out of the text-only version of the newsletter. This is basically so we can reuse
+ * the boilerplate, which contains HTML, but also cleans up unintended whitespace.
+ *
+ * @return {String} The text template cleaned of HTML
+ */
+export const cleanNewsletterTemplate = template => runSequence([
```
slifty: so incredibly clean

src/scripts/sendTestNewsletter.js
```diff
@@ -1,10 +1,23 @@
 import logger from '../server/utils/logger'
 
-import { HelloWorldNewsletter } from '../server/newsletters'
+import {
+  HelloWorldNewsletter,
+  PrimaryNewsletter,
```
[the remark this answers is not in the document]
slifty: Went with NationalNewsletter for now since it's fairly US-centric already and we'll probably refactor newsletter types before expanding to other countries anyway.

src/server/newsletters/AbstractNewsletter.js
```diff
@@ -195,6 +232,10 @@ class AbstractNewsletter {
    * @return {Promise} The promise generated by the Mailer's send()
    */
   send = async () => {
+    const isNewsletterSendable = await this.getIsNewsletterSendable()
+    if (!isNewsletterSendable) {
+      return Promise.reject(new Error('The newsletter did not satisfy its sendability requirements.'))
```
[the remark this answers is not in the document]
slifty: I'm going with `assertNewsletterIsSendable` -- also [assert](https://nodejs.org/api/assert.html) is a thing!  We can use that.

src/server/newsletters/templates/primary.hbs
slifty: Issue created: #96

src/server/newsletters/PrimaryNewsletter.js
slifty: Made this an issue to further discuss in #97

src/server/newsletters/PrimaryNewsletter.js
```diff
@@ -0,0 +1,45 @@
...
+  getMailingList = () => MAILING_LISTS.PRIMARY
+
+  getPathToTemplate = () => `${__dirname}/templates/primary.hbs`
+
+  getPathToTextTemplate = () => `${__dirname}/templates/primaryText.hbs`
+
+  getSubject = () => 'Tech & Check Alerts'
+
+  getIsNewsletterSendable = async () => {
+    const bodyData = await this.getCachedBodyData()
```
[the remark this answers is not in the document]
slifty: As noted elsewhere, moved this to #97

slifty:
Looks like the mailing list sends -- no names or affiliations but that's because they aren't being populated!

src/server/newsletters/AbstractNewsletter.js
```diff
@@ -126,6 +148,21 @@ class AbstractNewsletter {
     return mailingListAddress
   }
 
+  /**
+   * Provides the recipient mailing list address for the newsletter. To make sure we don't send
+   * newsletters to real mailing lists during development, this returns the developer mailing list
+   * if we're not in production mode or trying to send to one of our other interal mailing lists,
+   * like testers.
+   *
+   * @return {String} The intended newsletter recipient, or the developers mailing list
```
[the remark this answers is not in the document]
reefdog: I was mapping it (perhaps too directly) to [the message data object](https://github.com/TechAndCheck/tech-and-check-alerts/blob/967ec5f3602b1ea99410cad675dd773d83efe318/src/server/newsletters/AbstractNewsletter.js#L221-L228), where `subject` is populated by `getSubject()`, `recipient` by `getRecipient`, and `body` by a pair of functions (for text and HTML). And technically—although this would [break validation currently](https://github.com/TechAndCheck/tech-and-check-alerts/blob/967ec5f3602b1ea99410cad675dd773d83efe318/src/server/utils/mailer.js#L7)—the recipient could be in the form of `First Last <email@address.com>` since we end up [sticking it in the `to` Mailgun API field](https://github.com/TechAndCheck/tech-and-check-alerts/blob/967ec5f3602b1ea99410cad675dd773d83efe318/src/server/workers/mailer/Mailer.js#L28), which [accepts that format](https://documentation.mailgun.com/en/latest/api-sending.html#sending).

In effect: right now this is expected to be an email address and we have no plans for newsletters to send to the mailer with the `Name <email>` format, so I don't mind renaming it `getRecipientEmail()`. But perhaps the more forward-thinking thing is to allow for that format by documenting an example and fixing the validation.

Issue #90, opened 2019-07-26
Test email template in target clients and fix problems


reefdog:
We want to verify that the new email template (#89) will work on all the client email programs. @cguess has started a [Google Sheet tracking various clients](https://docs.google.com/spreadsheets/d/1-S_OnRw-xwN5q7mMSJvrZTqVR_f6bcL3ee9JQdoi2bs/edit#gid=0); let's make sure we hit a baseline of readability and conformity across all major clients.

reefdog:
This is looking good in target email clients. Closing it and will open client-specific issues as they're discovered.

PR #108, opened 2019-08-09
Fix newsletter rendering problems (thus far)

### LINKED ISSUE #96: Blockquotes render as bubbles in FastMail web client
Our newsletter looks so cool #92 but it is rendering a bit differently in clients that stylize their blockquotes.

image (8)

Lets see if we can't override that styling in the name of a consistent viewing experience.

reefdog:
So far, we've run into two classes of problems when testing the newsletter in various email clients:

1. CSS selectors are inconsistently applied across clients.
2. Some clients apply their own bespoke styling to certain elements.

The common resolution to (1) is to inline styles directly on elements. This results in massive redundancy, because it converts this:

```html
<style>li { list-style-type: none; }</style>
<ul>
  <li>One</li>
  <li>Two</li>
  <li>Three</li>
</ul>
```

To this:

```html
<ul>
  <li style="list-style-type: none;">One</li>
  <li style="list-style-type: none;">Two</li>
  <li style="list-style-type: none;">Three</li>
</ul>
```

However, it's also effective, and resolves a great number of email client styling problems. So, I've added the [Juice library](https://github.com/Automattic/juice#readme) and used it to inline CSS directly on elements while rendering the template. (Pseudo-elements and media queries are dynamic and thus can't be inlined; they're left alone, and we shouldn't use them for critical styles.)

One downside to Juice: evidently it depends on an earlier version of Cheerio, 0.22.0, than the 1.0.0-rc.3 we've been using. I haven't yet noticed a problem; tests and the scraper still seem to run fine. But if this is a blocker, we can try to sort it out!

Back to the list above: the solution to (2) is to snipe these as we find them. Thus, I've also created a section for these client-specific overrides/resets, and sniped the ugly Fastmail blockquotes as our first target.

Affects #90 
Fixes #96


src/server/newsletters/AbstractNewsletter.js
```diff
@@ -8,6 +8,7 @@ import {
 import {
   getHandlebarsTemplate,
   cleanNewsletterTemplate,
+  inlineTemplateStyles,
```
slifty: This name is a noun but it should be a verb -- e.g. "insertInlineStyles" or "convertStylesToInline" or "generateInlineStyleTemplate"

reefdog: I actually was verbing it, yes. "Inlining styles" is a thing I've said unfortunately often. But if that's confusing, I'd probably go with `moveStylesInline`.

slifty: Loooo oh right it is a verb too

Yeah if it takes 5 seconds lets change to moveStylesInline

If it takes 6 or more seconds we can leave it as is I GUESS.

reefdog:
Yeah, I glanced at [inline-css](https://www.npmjs.com/package/inline-css), but it's quite new and Juice is supported by Auttomattic with substantially more activity. And besides, it's package-lock.json points to the same version of Cheerio, so wouldn't even resolve that.

> Doesn't NPM handle sub-dependency crossover issues so we don't have to worry about that?

Probably/hopefully? Just wanted to pre-note it. In practice doesn't seem to be an issue.

src/server/newsletters/AbstractNewsletter.js
reefdog: It took two minutes.

Issue #64, opened 2019-06-15
DRY out the regular expressions in scraper tools


slifty:
The cnn scraper tools have a few repeat regular expressions, specifically around attribution, affiliation, and speaker names.  We should replace the `/[A-Z]*/` style expression definitions with `RegEx()` definitions which will allow us to use variables (and therefore reuse sub components).

slifty:
I actually... think we can probably close this out.  With the recent fixes related to #167 I believe the most egregious duplicate segments have been removed.  In the cases where we still have somewhat duplicate overlap (e.g. `attributionRegex` vs `attributionAffiliationRegex`) it is intentional / necessary.

PR #112, opened 2019-08-12
Allow for numbers in speaker affiliations

### LINKED ISSUE #98: Attributions can contain numbers
```
BILL CLINTON, 42ND PRESIDENT OF THE UNITED STATES: All of a sudden 25 million more people had healthcare and then the people are out there busting it, sometimes 60 hours a week wind up with their premiums doubled and their coverage cut in half.
```

We need the attribution regular expression to allow for numbers.

slifty:
This adds tests and support for speakers who have numbers in their affiliations.

Resolves #98

### DIFF
--- a/src/server/utils/cnn.js
+++ b/src/server/utils/cnn.js
@@ -10,6 +10,13 @@
  * @property {String}  text    What the person actually said
  */
 
+/**
+ * These expressions are used in various utility mehtods
+ * to identify special portions of scraped tex.
+ */
+const attributionaAffiliationRegex = /,([A-Z\d\s"(),]*)/
+const chunkAttributionRegex = /[A-Z\d\s"(),.]+[:]/
+
 export const isTranscriptListUrl = url => url.startsWith('/TRANSCRIPTS/')
   && url.endsWith('.html')
 
@@ -76,9 +83,6 @@ export const addBreaksOnSpeakerChange = transcript => transcript
  */
 export const splitTranscriptIntoChunks = transcript => transcript.split('\n')
 
-// Used to identify the speaker section of a chunk
-const chunkAttributionRegex = /[.,A-Z\s"()]+[:]/
-
 /**
  * Extract the speaker name / affiliation from a chunk.
  *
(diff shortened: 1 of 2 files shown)


src/server/utils/cnn.js
```diff
@@ -10,6 +10,15 @@
...
  */
 
+/**
+ * These expressions are used in various utility mehtods
+ * to identify special portions of scraped tex.
+ */
+// Used to identify the affiliation section of an attribution
+const attributionaAffiliationRegex = /,([\dA-Z\s,"()]*)/
+// Used to identify the speaker section of a chunk
+const chunkAttributionRegex = /[.,A-Z\s"()\d]+[:]/
```
reefdog: I know #64 will improve this substantially (hopefully) but until then, could we at least rearrange the rules within the `[]` for easier comparison of the common parts? E.g.:

```js
const attributionaAffiliationRegex = /,([A-Z\d\s"(),]*)/
const chunkAttributionRegex = /[A-Z\d\s"(),.]+[:]/

Comment (slifty) [reply src/server/utils/cnn.js]: Sure thing -- also I honestly am wondering if we really need to list out  all of the possibilities...  Honestly, the only important things are:

"attribution ends in a :"

and

"affiliation comes after the first , of an attribution"

Maybe I should just... use a `.*`

Comment (slifty) [reply src/server/utils/cnn.js]: Update:

This passes our current tests:

```
// Used to identify the affiliation section of an attribution
const attributionaAffiliationRegex = /,(.*)/
// Used to identify the speaker section of a chunk
const chunkAttributionRegex = /.+[:]/
```

I'm thinking I should just move it to this and we can move on with our lives...

reefdog: That's nicely succinct but looks like it could result in false positives. We should add some tests specifically to surface false positives and see if that works!

slifty: Oh right, that's why.

OK so really the "trick" is that attributions are upper case.   `.*` definitely won't work, but `[^a-z]*` might.

I'll just put this in an issue and move along with this with the change you requested.

slifty: Issue opened: #113

PR #114, opened 2019-08-12
Switch attribution regex to negatives

### LINKED ISSUE #107: Names can contain apostrophes and hyphens
- [Apostrophe example](http://cnn.com/TRANSCRIPTS/1907/31/ip.01.html): `BETO O'ROURKE (D), PRESIDENTIAL CANDIDATE`. What've you got against the Irish, @slifty??
- [Hyphen example](http://www.cnn.com/TRANSCRIPTS/1907/31/ip.01.html): `GOV. STEVE BULLOCK (D-MT), PRESIDENTIAL CANDIDATE`. What've you got against states, @slifty?? (Edit: Wait, maybe we filter out `(D-MT)` anyway?)

For reference: [Asa's existing method for ID'ing speakers](https://github.com/TechAndCheck/tech-and-check-alerts-2018/blob/ab1bb599319f034dbe300d2890b1638df764d89e/cnn_tools/cnn_buster.py#L62-L101).

### LINKED ISSUE #113: Consider a more universal speaker / attribution regex
#98, #107, and future issues are around bugs that crop up when attributions have characters we didn't predict.

Rather than playing whack-a-mole it would be great to have a more universal regex expression.  The key is not generating false positives (which is why `.*:` wouldn't work).  However, ultimately the issue here is that affiliations NEVER have lower case letters.  It seems like they can, in theory, have any other kind of character.

Given that trick, lets think about `[^a-z]*:` as an alternative to the list of valid characters.

This will have to be paired with some tests.

slifty:
Instead of listing all of the things an attribution can contain, lets just list out all of the things it can't.  The simple rule is that attributions never have lower case letters.

Attribution, name, and affiliation now use this logic instead.  If we find false positives cropping up then we can move from there.

Resolves #113 
Resolves #107

### DIFF
--- a/src/server/utils/cnn.js
+++ b/src/server/utils/cnn.js
@@ -11,10 +11,11 @@
  */
 
 /**
- * These expressions are used in various utility mehtods
+ * These expressions are used in various utility methods
  * to identify special portions of scraped tex.
  */
 const attributionAffiliationRegex = /,([^a-z]*)/
+const attributionNameRegex = /[^a-z,]+/
 const chunkAttributionRegex = /[^a-z]+[:]/
 
 export const isTranscriptListUrl = url => url.startsWith('/TRANSCRIPTS/')
@@ -111,7 +112,7 @@ export const getStatementFromChunk = chunk => chunk
  * @return {String}             The portion of the attribution that contains the person's name
  */
 export const getNameFromAttribution = attribution => (
-  (attribution.match(/[A-Z\s.()]+/)
+  (attribution.match(attributionNameRegex)
   || [''])[0]
 )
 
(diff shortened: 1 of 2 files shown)


src/server/utils/cnn.js
```diff
@@ -14,8 +14,9 @@
  * These expressions are used in various utility mehtods
```
reefdog: ugh `mehtods` slipped into an old commit uncaught, sorry.

slifty: Will fix here.

slifty:
I'm running in practice and not thrilled -- though not because this made it worse, just because of the bugs I just uncovered around (parenthetical descriptions in affiliations).

I'll merge, then do another PR with that.  We'll get where we need to get gosh darn it.

PR #122, opened 2019-08-12
Add a scraper test utility

### LINKED ISSUE #120: Add ability to test statement scrapes for a given URL
When debugging I want to know what the scraper extracts for a given URL without having to re-run the full scraper and wait.  Lets add a yarn utility method for that.

slifty:
You can now call `yarn scraper:cnn -u /TRANSCRIPT/blahblahblah.html` and it will run the statement scraper on that page.

Results get output to INFO console.

Resolves #120

### DIFF
--- a/src/scripts/scrapeCnnStatements.js
+++ b/src/scripts/scrapeCnnStatements.js
@@ -1,0 +1,31 @@
+import CnnTranscriptStatementScraper from '../server/workers/scrapers/CnnTranscriptStatementScraper'
+import { isTranscriptUrl } from '../server/utils/cnn'
+import logger from '../server/utils/logger'
+
+const getUrl = () => {
+  const urlFlag = process.argv.indexOf('-u')
+  if (urlFlag !== -1) {
+    const url = process.argv[urlFlag + 1]
+    if (isTranscriptUrl(url)) return url
+  }
+  return null
+}
+
+const url = getUrl()
+if (!url) {
+  logger.error('Pass in a valid CNN transcript URL with the `-u` parameter.')
+  process.exit()
+}
+
+const scrapeStatements = () => {
+  const scraper = new CnnTranscriptStatementScraper(url)
+  return scraper.run()
+}
+
+scrapeStatements().catch((error) => {
+  logger.error(`Scraping failed. ${error}`)
+}).then((statements) => {
+  logger.info(`Total statements: ${statements.length}`)
+  statements.forEach(statement => logger.info('%o', statement))
+  process.exit()
+})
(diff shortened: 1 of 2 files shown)


src/scripts/scrapeCnnStatements.js
```diff
@@ -0,0 +1,31 @@
+import CnnTranscriptStatementScraper from '../server/workers/scrapers/CnnTranscriptStatementScraper'
+import { isTranscriptUrl } from '../server/utils/cnn'
+import logger from '../server/utils/logger'
+
+const getUrl = () => {
+  const urlFlag = process.argv.indexOf('-u')
+  if (urlFlag !== -1) {
```
reefdog: nbd and you need this position anyway, _but_ I do think we should try to use `includes()` rather than the old-fashioned check for `-1`. It's just so much cleaner.

So this could be:
```js
const URL_FLAG = '-u'
if (process.argv.includes(URL_FLAG) {
  const url = process.argv[process.argv.indexOf(URL_FLAG) + 1]
  // …
}
```
Which, yeah, is more verbose.

src/scripts/scrapeCnnStatements.js
```diff
@@ -0,0 +1,31 @@
...
+    const url = process.argv[urlFlag + 1]
+    if (isTranscriptUrl(url)) return url
```
reefdog: If we do use `-u` but fail to actually pass in a value, e.g. `yarn scraper:cnn -u`, we don't actually catch that case and instead trigger a TypeError:

```
const isTranscriptUrl = url => url.startsWith('/TRANSCRIPTS/') && url.endsWith('.html');
                                   ^
TypeError: Cannot read property 'startsWith' of undefined
```
Again, nbd, but it'd be nice to either check for this before passing to `isTranscriptUrl`, or wrap in a try/catch, or something.

src/scripts/scrapeCnnStatements.js
slifty: I DO NOT KNOW WHAT I AM SUPPOSED TO DO NOW THANK YOU JUSTIN.

reefdog: You're supposed to make things perfect the first time just the way I like them.

reefdog: (My basic critique was "eww now that we have `includes()` let's pretend `=== -1` never existed.) (But it's not a merge-blocking critique.)

slifty:
(Merging to keep momentum, but feeling bad about myself, so, win win for you!)

PR #140, opened 2019-08-13
Clean affiliation names more effectively

### LINKED ISSUE #135: Party affiliation throws off normalization
Noticed a lot of politicians not being properly normalized in our claims lists.  [Here's an example transcript](http://www.cnn.com/TRANSCRIPTS/1908/04/se.03.html)

Here's an example candidate intro:

```
SEN. ELIZABETH WARREN (D-MA), PRESIDENTIAL CANDIDATE: I do and if you know you think about the sorrow that's just mounting in this country. Every one of those people who died is somebody's sister, somebody's brother, somebody's child, somebody's parent, our friends, our neighbors and all the people who were wounded. 
```
This would parse the association as follows:

```
{
  name: "SEN. ELIZABETH WARREN (D-MA)",
  affiliation: "PRESIDENTIAL CANDIDATE:"
}
```

This causes a normalization problem because normalization is currently fairly simple -- it looks at the END of strings.  The (D-MA) is not actually part of the name, so when `WARREN` appears later it doesn't make the connection.

We should either:

1. Remove the (D-MA) entirely from names and never show it
2. Remove the (D-MA) temporarily during the normalization process

I'm leaning to (1) for now, since I don't think we'll want to show (D-MA) in the newsletter and it really isn't part of the person's name; it's a separate attribute that we aren't currently set up to track.

slifty:
Parentheticals in affiliation names were not being removed, and furthermore hyphens in affiliations generally were causing chunks to be improperly split.  This corrects both of those problems.

Ultimately we need to DRY out the regular expressions as described in issue #64.  I have a feeling that edge cases aren't being fully corrected as we expand what an affiliation can look like.

Resolves #135

### DIFF
--- a/src/server/utils/__test__/cnn.test.js
+++ b/src/server/utils/__test__/cnn.test.js
@@ -158,6 +158,10 @@ describe('utils/cnn', () => {
       expect(addBreaksOnSpeakerChange('"This is a quote." DONNA: Whoever it is they seem to be running out of ideas.'))
         .toBe('"This is a quote."\nDONNA: Whoever it is they seem to be running out of ideas.')
     })
+    it('Should allow for hyphens in affiliations', () => {
+      expect(addBreaksOnSpeakerChange('Testing this. SEN. ELIZABETH WARREN (D-MA), PRESIDENTIAL CANDIDATE: I do and if you know you think about the sorrow'))
+        .toBe('Testing this.\nSEN. ELIZABETH WARREN (D-MA), PRESIDENTIAL CANDIDATE: I do and if you know you think about the sorrow')
+    })
   })
 
   describe('splitTranscriptIntoChunks', () => {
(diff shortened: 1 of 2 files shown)


src/server/utils/__test__/cnn.test.js
```diff
@@ -346,6 +350,12 @@ describe('utils/cnn', () => {
       expect(cleanSpeakerName('REPRESENTATIVE JIMMY DUST'))
         .toBe('JIMMY DUST')
     })
+    it('Should remove parentheitcals', () => {
```
reefdog: `parentheitcals` → `parentheticals` (which also isn't a word, but is a properly spelled not-word)

src/server/utils/__test__/cnn.test.js
```diff
@@ -346,6 +350,12 @@ describe('utils/cnn', () => {
...
+      expect(cleanSpeakerName('JIMMY DUST (D-MA)'))
+        .toBe('JIMMY DUST')
+      expect(cleanSpeakerName('JIMMY DUST (THE MOST NOTORIOUS BOTANIST IN THE WEST)'))
```
reefdog: A flagrant exaggeration.

PR #144, opened 2019-08-13
Allow envs to set a global mailing list override

### LINKED ISSUE #138: Allow global override of newsletter destinations
Currently, newsletters are individually [associated with a specific mailing list](https://github.com/TechAndCheck/tech-and-check-alerts/blob/3fd849a9f923fbc3868fa836c4a16dfc36877a69/src/server/newsletters/NationalNewsletter.js#L12). Then, while sending, we check the current environment; if we're in production or if the newsletter is already configured to go to an internal mailing list (devs or testers), we send it to its configured destination. Otherwise, we send it to the dev mailing list.

That model is fine for normal times, but there are times — such as while setting up a new production environment — that we may want to run the app in production mode (connecting to the production database, etc.), but not have newsletters sent to the production recipients. Thus it would be nice to be able to set an env variable that said "send all newsletters to this address" and have that override all other logic.

Of course, with much power comes much responsibility (to remember that we've set this).

@slifty Do you have any thoughts on this before I just go the .env route? It's really just guardrails during specific times, so I think that's the right way?

reefdog:
We gain a new .env example (`MAILING_LIST_OVERRIDE `) that, when configured, replaces the newsletter-specific mailing list configuration. This mostly exists to let us setup otherwise-production environments that are globally configured to send only to internal mailing lists until otherwise let off the leash to send to their true configured mailing lists.

NB: This still has the old guards that won’t let us accidentally configure local development environments to send to the real lists. (Or, it shouldn’t. I’m scared to test.) If you ever want to do that, you will have to manually get in there and remove the guards and may God have mercy on your soul.

This also moves the function that composes mailing list email addresses off of `AbstractNewsletter` and to a utility function where it belongs, plus tests.

Resolves #138

### DIFF
--- a/src/server/newsletters/AbstractNewsletter.js
+++ b/src/server/newsletters/AbstractNewsletter.js
@@ -1,11 +1,8 @@
 import config from '../config'
+import { getFileContents } from '../utils'
 import {
-  isProductionEnv,
-  getFileContents,
-} from '../utils'
-import {
-  isInternalMailingList,
   getMailingListAddress,
+  guardMailingList,
 } from '../utils/newsletters'
 import {
   getHandlebarsTemplate,
@@ -14,7 +11,6 @@ import {
 } from '../utils/templates'
 import Mailer from '../workers/mailer'
 import logger from '../utils/logger'
-import { MAILING_LISTS } from './constants'
 
 class AbstractNewsletter {
   constructor() {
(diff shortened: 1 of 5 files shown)


src/server/newsletters/AbstractNewsletter.js
```diff
@@ -138,32 +140,25 @@ class AbstractNewsletter {
...
-   * @return {String} The intended newsletter recipient, or the developers mailing list
+   * @return {String} The intended newsletter recipient
    */
   getRecipient = () => {
-    const mailingList = this.getMailingList()
-    return (isProductionEnv() || isInternalMailingList(mailingList))
-      ? this.getMailingListAddress()
-      : MAILING_LIST_ADDRESSES[MAILING_LISTS.DEVELOPERS]
+    const configuredMailingList = config.MAILING_LIST_OVERRIDE || this.getMailingList()
+    const guardedMailingList = (isProductionEnv() || isInternalMailingList(configuredMailingList))
```
slifty: Seeing this in isolation makes me realize this should be a named utility method.

reefdog:
Agreed! Moved it to a utility method and added comments/tests. Can you give it a quick re-review?

PR #146, opened 2019-08-13
Increase newsletter claim limit to 15

### LINKED ISSUE #133: Increase claim limit to 15
Current newsletter chooses top 15 claims of each type. Do the same.

reefdog:
The old value of 10 was a shot in the dark on my part. Changing to 15 to match current newsletter.

BTW I'm pretty sure this will soon be a constant, but I'm only planning to do that if 15 is our common limit (i.e. once we add other claim types / newsletters), not just to get limit literals out of the class definition.

Resolves #133

### DIFF
--- a/src/server/newsletters/NationalNewsletter.js
+++ b/src/server/newsletters/NationalNewsletter.js
@@ -12,7 +12,7 @@ import {
 const { Claim } = models
 
 class NationalNewsletter extends AbstractNewsletter {
-  getMailingList = () => MAILING_LISTS.PRIMARY
+  getMailingList = () => MAILING_LISTS.NATIONAL
 
   getPathToTemplate = () => `${__dirname}/templates/national.hbs`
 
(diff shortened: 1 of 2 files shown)


reefdog:
BTW I'm pretty sure this will soon be a constant, but I'm only planning to do that if 15 is our common limit (i.e. once we add other claim types / newsletters), not just to get limit literals out of the class definition.

src/server/newsletters/NationalNewsletter.js
```diff
@@ -23,7 +23,7 @@ class NationalNewsletter extends AbstractNewsletter {
   }
 
   fetchTVClaims = () => (Claim.findAll({
-    limit: 10,
+    limit: 15,
```
slifty: I do hate to say this BUT... you're right that this should be a constant.

e.g.

```
NEWSLETTER_SETTINGS = {
   NATIONAL: {
     ...
   }
}
```

reefdog: How dare you accuse me of being right.

Issue #129, opened 2019-08-13
Our transcript URL validation methods actually validate paths


reefdog:
We have a [pair of methods](https://github.com/TechAndCheck/tech-and-check-alerts/blob/db55c18ccaf0ea688848b43c59b109944179117b/src/server/utils/cnn.js#L21-L25) that nominally validate the format of a transcript URL, but actually expect to receive a path.

We should modify these to expect fully-qualified URLs. If we have a need to truly check a path, then they should be renamed `isTranscriptPath`/`isTranscriptListPath`, as they are not URLs.

(This also addresses [my complaint on this PR](https://github.com/TechAndCheck/tech-and-check-alerts/pull/122#pullrequestreview-274371208).)

slifty:
Pedantic question: is a relative URL not technically a URL?  These two methods are often called in the context of scraping a relative url from an anchor tag on a page.

The main question is:  should it reject relative URL paths or should it simply accept absolute ones?

That said, we do ultimately create absolute paths at some point before triggering a crawl, so it's fine to require it to happen before it's considered valid.

reefdog:
Pretty sure that technically and pedantically a relative URL isn't a URL. It's a relative URL.

I don't mind if we change the validation functions to pass the parameter through `getFullCnnUrl()` before checking them (and checking them against `http://cnn.com/TRANSCRIPTS/`), thus accepting either absolute or relative. But we should document that either is allowed, since we're now being a little casual with our use of "URL".

slifty:
I'm totally fine demanding they be real URLs I just wanted to get that clarification -- fully agree with the decision to change this to actually validating URLs

reefdog:
Stashing some WIP for the util tests:

```js
const urls = {
  transcriptPortal: [
    'http://transcript.cnn.com/TRANSCRIPTS/',
  ],
  transcriptList: [
    'http://transcripts.cnn.com/TRANSCRIPTS/2019.08.12.html',
  ],
  transcript: [
    'http://transcript.cnn.com/TRANSCRIPTS/1908/18/cnr.20.html',
  ],
  google: 'http://google.com',
}
```

PR #148, opened 2019-08-14
Add date cutoff parameter

### LINKED ISSUE #136: Add support for relative date scrape cutoffs
Right now the CNN scraper goes back in time until there is nothing left to scrape.  This isn't actually useful, since we don't care about old claims (and in fact likely don't want to store them at all as we are discussing in #37).

We need to be able to configure the system to only look back a certain number of days when scraping.  For now I think we can have this be universal across all statement scrapers (as opposed to configuring each scraper cutoff separately).  It should be a relative date (e.g. "past 5 days") so that as time marches on the scraper stops scraping things from the past even if we wipe scrape longs, etc.

slifty:
This creates a new environment variable SCRAPE_DAY_HORIZON which specifies the number of days prior to the current date that should be scraped for claims.

It also adds a few tests and utility methods that can be used in other scraper pipelines eventually.

This continues the whole "call a URL a path" thing from #129 in the new `extractPublicationDateFromUrl` method.  We'll need to fix it here as well when we fix that issue.

Right now there is no way to say "scrape all known claims" since there's no real use case for that.

Resolves #136


src/server/models/index.js
```diff
@@ -10,7 +11,13 @@ const env = config.NODE_ENV || ENV_NAMES.DEVELOPMENT
 const envConfig = sequelizeConfig[env]
 
 const db = {}
-const sequelize = new Sequelize(envConfig.url, envConfig)
+const sequelize = new Sequelize(
+  envConfig.url,
+  Object.assign(
+    { logging: msg => logger.debug(msg) },
```
reefdog: :squint: Why aren't we doing this in `sequelize-config.js`?

src/server/utils/scraper.js
```diff
@@ -0,0 +1,8 @@
+// Disabling because we intend to have more exports in the future.
+/* eslint-disable import/prefer-default-export */
+import dayjs from 'dayjs'
+import config from '../config'
+
+export const isDateBeforeScrapeHorizon = date => dayjs()
```
reefdog: My literary bent wants this to be called `isDateBeyondScrapeHorizon` for metaphorical consistency but this is fine and correct.

In any case does the `date` param have to be a DayJS object, or can it be a DayJS-parseable string? Worth documenting this function to make this clear.

src/server/queues/cnnTranscriptListCrawlerQueue/cnnTranscriptListCrawlerJobProcessor.js
```diff
@@ -10,9 +12,11 @@ const scrapeTranscriptUrl = url => statementScraperQueue.add({ url })
 const processTranscriptUrl = async (url) => {
   const scraper = new CnnTranscriptStatementScraper(url)
   const recentScrapeTime = await scraper.getMostRecentScrapeTime()
-
+  const urlPublicationDate = extractPublicationDateFromTranscriptUrl(url)
```
reefdog: A neat shortcut! So much faster than having to actually scrape the transcript and parse out its date there. Hope they never stop using date-embedded URLs…

src/server/utils/__test__/cnn.test.js
```diff
@@ -68,6 +69,24 @@ describe('utils/cnn', () => {
...
+    it('Should extract dates', () => {
+      const date = extractPublicationDateFromTranscriptUrl('/TRANSCRIPTS/1906/01/cnr.20.html')
+      expect(date)
+        .not.toBeNull()
+      expect(date.date())
+        .toEqual(1)
+      expect(date.year())
+        .toEqual(2019)
+      expect(date.month())
+        .toEqual(5)
```
reefdog: This is all fine, but L77-82 could be shortened to `expect(date.format('YYYY-MM-DD').toBe('2019-06-01')`.

src/server/utils/__test__/scraper.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+describe('isDateBeforeScrapeHorizon', () => {
+  const day = dayjs()
+  it('Should recognize dates before the date horizon', () => {
+    const earlyDate = day.subtract(config.SCRAPE_DAY_HORIZON + 1, 'day')
+    expect(isDateBeforeScrapeHorizon(earlyDate))
+      .toBe(true)
+  })
+  it('Should recognize dates after the date horizon', () => {
+    const earlyDate = day.add(1, 'day')
```
reefdog: Rename `earlyDate` to `validDate` or something, yeah?

src/server/utils/cnn.js
```diff
@@ -30,6 +35,25 @@ export const getFullCnnUrl = (url) => {
   return `http://cnn.com/${url}`
 }
 
+/**
+ * Pull out the publication date from a passed transcript url.
```
reefdog: `url` → `URL`

src/server/utils/cnn.js
```diff
@@ -30,6 +35,25 @@ export const getFullCnnUrl = (url) => {
...
+ *
+ * If the url is not a valid transcript URL, throw an error.
```
reefdog: `If the url` → `If the URL`

src/server/models/index.js
slifty: That isn't babel'd so we can't import the logger there.

reefdog: 👀 ah ok.

src/server/utils/cnn.js
reefdog: Actually gonna do this as a hotfix to master to grab the rest.

src/server/utils/cnn.js
reefdog: Actually gonna do this as a hotfix to master to grab the rest.

PR #151, opened 2019-08-14
Add national newsletter queue

### LINKED ISSUE #134: Add national newsletter queue
Create a queue for sending the national newsletter, configured to send every morning.

reefdog:
This is it. This adds a national newsletter queue, scheduled every day at 9am (but in what timezone oooOOOoohhHHH?) (no but really, what timezone? will this be the deployed system's local time?)

Resolves #134


reefdog:
For sure. I was actually thinking less complicated than that: just that we should determine the server timezone and offset this schedule appropriately (since I think we want it sending at 7am Eastern?) But I'll open a new issue for that.

Issue #162, opened 2019-08-16
Running master on EC2 has issues


reefdog:
There's a lot of Slack backscroll behind this, but in summary: scheduling the current tasklist with `yarn queue:jobs:schedule` on our current EC2 setup triggers a host of issues, some of which could be environmental and some configuration. We scaled the instance up to `t2.large` for further testing (which gave us the RAM headroom particularly), though I scaled it back to `t2.small` before calling it a day. You'll want to re-scale-up (and maybe even add some more cores if you think CPU might be an issue).

1. **Bull:** `error.log` shows a ton of `Missing lock for job` entries. Perhaps [this section from the Bull docs](https://github.com/OptimalBits/bull#important-notes) would be useful.
2. **Redis:** `error.log` shows a lot of `Redis is configured to save RDB snapshots, but it is currently not able to persist on disk` entries
3. **Sequelize/Postgres:** The console shows huge batches of Sequelize connection timeouts.
4. **Promises:** The console shows a whole bunch of complaints about unresolved async functions from the Sequelize timeouts in (3), which really just means #119 is a thing.

If I'm watching things with `htop`, the CPU immediately pegs both cores at 100%, and RAM usage gradually increases until settling in around 5GB. Sometimes CPU use settles down to 0% for a while, but RAM usage never seemed to drop below ~3GB. The console would burst tons of those Sequelize timeout messages (and associated unresolved Promise complaints) periodically.

reefdog:
BTW, I changed the schedule of the `CnnTranscriptPortalCrawlerJobScheduler` to `EVERY_HOUR` on the server (not committed to the repo) while testing this.

slifty:
It took a bunch of digging but I believe I discovered the root cause.

The issue stems from a misunderstanding about the way Bull handled concurrency.  In short: Bull creates child processes equal to the concurrency setting EVERY TIME you run .process on a queue object.  Because of the way that our QueueFactory worked we were doing that every time a queue was requested (even if the only purposes was to insert new jobs).

A PR is incoming to fix this.

reefdog:
We can close this now, yeah? Yeah?

slifty:
Lets wait until it really runs but I think so.

reefdog:
lol fair. Offense, false start. Five-yard penalty, repeat third down.

slifty:
Looks like we're all set!  Yay!

PR #168, opened 2019-08-20
Split getQueue into two methods

slifty:
I spent a lot of time digging around and documented that outcome in [this potentially relevant bull issue](https://github.com/OptimalBits/bull/issues/923#issuecomment-522665737).

We had misunderstood Bull's concurrency mechanism, and whenever we generated a new queue we would create a queue with the processor attached.  This actually meant that every single time we ran getQueue() we were also creating a new job processor for that queue.

This meant that instead of having one job processor per queue we were having... many more.

This commit creates two types of Queue factory method:

1. getInsertionQueue -- this creates a queue objec t solely for the purpose of inserting new jobs.  Generally speaking this is the method that will be called.

2. getProcessingQueue -- this creates a queue object that is registered to process jobs.  This should never be called from inside a queue processor script and should really only be called once.

This commit also changes the template for QueueScheduler objects -- it didn't make sense for them to each implement an external getQueueFactory method and then invoke it identically to override an internal getQueue.

We now have getQueueFactory as the abstract method and getQueue sitting as an implemented method in the Abstract class.

This is related to #162 though we won't know for sure that it resolves until after deploying.

### DIFF
--- a/src/server/queues/AbstractJobScheduler.js
+++ b/src/server/queues/AbstractJobScheduler.js
@@ -14,16 +14,22 @@ class AbstractJobScheduler {
   }
 
   /**
-   * Abstract method that provides the queue object that the schedule
+   * Abstract method that provides the queue factory for the queue that the schedule
    * should be applied to.
    *
    * OVERRIDE WHEN EXTENDING
    *
-   * @return {Queue} the queue object
+   * @return {AbstractQueueFactory} The instantiated queue factory
    */
-  getQueue = () => {
-    throw new Error('You extended AbstractJobScheduler but forgot to define getQueue()')
+  getQueueFactory = () => {
+    throw new Error('You extended AbstractJobScheduler but forgot to define getQueueFactory()')
   }
+
+  /**
+   * Invokes getQueueFactory to actually get the insertion queue
+   * @return {Queue} The queue to schedule against
+   */
+  getQueue = () => this.getQueueFactory().getQueue()
 
   /**
    * Method that provides the job information that should be part of the scheduled
--- a/src/server/queues/AbstractQueueFactory.js
+++ b/src/server/queues/AbstractQueueFactory.js
@@ -37,8 +37,13 @@ class AbstractQueueFactory {
     )
     queue.on('error', error => logger.error(error))
     queue.on('failed', (job, error) => logger.warn(error))
-    queue.process(this.getPathToProcessor())
     return queue
+  }
+
+  startQueueProcessor = () => {
+    logger.debug(`Starting queue processor: ${this.getQueueName()}`)
+    const queue = this.getQueue()
+    queue.process(this.getPathToProcessor())
   }
 }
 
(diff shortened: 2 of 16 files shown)


src/server/queues/AbstractQueueFactory.js
```diff
@@ -30,13 +30,19 @@ class AbstractQueueFactory {
...
     const queue = new Queue(
       this.getQueueName(),
       config.REDIS_URL,
     )
     queue.on('error', error => logger.error(error))
     queue.on('failed', (job, error) => logger.warn(error))
+    return queue
+  }
+
+  getProcessingQueue = () => {
```
reefdog: I think `getProcessingQueue()` isn't a terribly good name for what's going on here. The aim of this function is to start the processor; the fact that it returns the queue at the end is a bit of convenience. I feel like it'd be clearer if this was renamed `startProcessingQueue()` or similar, and then returned nothing.

If we find ourselves commonly wanting to start the processor and then deal with the queue, we could create a utility function that did so: `getAndStartProcessingQueue()` or something.

Consequently, we may be able to rename `getInsertionQueue()` back to `getQueue()` since it doesn't need to be distinguished from `getProcessingQueue()`, because I think now that it doesn't run `process()` it acts in an expected way. But I'll leave that up to you!

(BTW for historians: this comment summarizes some Slack conversation with @slifty so he's to thank for some of the solutioneering.)

src/server/queues/AbstractQueueFactory.js
```diff
@@ -30,13 +30,19 @@ class AbstractQueueFactory {
     redis: config.REDIS_URL,
   })
 
-  getQueue = () => {
+  getInsertionQueue = () => {
```
reefdog: See comment on `getProcessingQueue()` about possibly renaming this back to `getQueue()`.

src/server/queues/AbstractQueueFactory.js
slifty: I'm only OK with this because it won't require me to rename the PR, as it still splits into two methods.

slifty: Oh: are you OK with `startQueueProcessing` as a name

reefdog: I really prefer `makeGetQueueMakeSetFactoryFartory`

slifty: Too bad, I made it startQueueProcessor

src/server/queues/AbstractJobScheduler.js
```diff
@@ -14,17 +14,23 @@ class AbstractJobScheduler {
...
+   * @return {AbstractQueueFactory} The instantiated queue factory
    */
-  getQueue = () => {
-    throw new Error('You extended AbstractJobScheduler but forgot to define getQueue()')
+  getQueueFactory = () => {
+    throw new Error('You extended AbstractJobScheduler but forgot to define getQueueFactory()')
   }
 
+  /**
+   * Invokes getQueueFactory to actually get the insertion queue
```
reefdog: Now that we aren't using `getInsertionQueue()` we can remove "insertion" from this sentence.

Issue #150, opened 2019-08-14
Unhandled promise rejection warning when scheduling jobs


reefdog:
When running `yarn queue:jobs:schedule`, the console dumps out the following before starting to log crawl/scrape progress:

```
Justins-Retina-MBP:tech-and-check-alerts justin$ yarn queue:jobs:schedule
yarn run v1.17.3
$ babel-node -- src/scripts/scheduleJobs
(node:60177) UnhandledPromiseRejectionWarning: Error: Invalid characters, got value: undefinedundefinedundefined
    at Function._parseField (/Users/justin/Projects/duke/tech-and-check-alerts/node_modules/cron-parser/lib/expression.js:186:11)
    at parse (/Users/justin/Projects/duke/tech-and-check-alerts/node_modules/cron-parser/lib/expression.js:774:36)
    at Function.parse (/Users/justin/Projects/duke/tech-and-check-alerts/node_modules/cron-parser/lib/expression.js:835:10)
    at Function.parseExpression (/Users/justin/Projects/duke/tech-and-check-alerts/node_modules/cron-parser/lib/parser.js:41:25)
    at getNextMillis (/Users/justin/Projects/duke/tech-and-check-alerts/node_modules/bull/lib/repeatable.js:202:29)
    at Queue.nextRepeatableJob (/Users/justin/Projects/duke/tech-and-check-alerts/node_modules/bull/lib/repeatable.js:33:24)
    at /Users/justin/Projects/duke/tech-and-check-alerts/node_modules/bull/lib/queue.js:688:19
    at async Promise.all (index 0)
(node:60177) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:60177) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

reefdog:
Actually, this is due to the `Schedules.None` typo in these three queues:

https://github.com/TechAndCheck/tech-and-check-alerts/blob/a7b63e1fe2e58049e95bc4acd6d50e0ca13a767e/src/server/queues/cnnTranscriptListCrawlerQueue/CnnTranscriptListCrawlerJobScheduler.js#L8
[...]

PR #176, opened 2019-08-20
Stop scheduling jobs that have a `NONE` schedule

### LINKED ISSUE #172: Stop scheduling unscheduled jobs
We have the capability to create job schedules that are basically no-ops: they follow the standard design, but have a [schedule of `NONE`](https://github.com/TechAndCheck/tech-and-check-alerts/blob/514db52584a4784a44efacb00979aaea577ab020/src/server/queues/constants.js#L6) and shouldn't actually be "scheduled". However, we don't actually pay attention to this; we blithely run `AbstractJobScheduler#scheduleJobs()` and [add the scheduler to the queue with a repeatables option that breaks Bull](https://github.com/TechAndCheck/tech-and-check-alerts/blob/514db52584a4784a44efacb00979aaea577ab020/src/server/queues/AbstractJobScheduler.js#L38-L46) (because it breaks the cron-parser; turns out `cron: 'none'` is meaningless).

We need to resolve this by recognizing the `NONE` schedule during `AbstractJobScheduler#scheduleJobs()` and skipping the `Queue#add()`. (I'll further guard against it by filtering down to only scheduleable queues when scheduling the jobs in the first place.)

reefdog:
We have the ability to create a job scheduler that should not actually be scheduled by configuring its schedule as `NONE`. Problem is, we still added a repeatable job to the queue, but with an invalid cron pattern. What we should’ve been doing is not schedule these jobs at all; `NONE` should make them a scheduling no-op.

This commit makes that real (jazz hands). Now the `scheduleJobs()` verifies the schedule isn’t `NONE` before scheduling it, else it returns `null`. (This does mean any code that checks the return value of `scheduleJobs()` should know that it may receive a `Job` object _or_ a `null` value. Thus the [`filter(job => !!job)` guardrail on the `scheduleJobs.js` task](https://github.com/TechAndCheck/tech-and-check-alerts/blob/b19aad2ff2b2cf70d00338a8282591d6e6e232a8/src/scripts/scheduleJobs.js#L7).)

I also modified the `scheduleJobs.js` task to prefilter jobs by those that even can be scheduled, by checking the `getIsScheduled()` method of each. Catch ‘em comin’ and goin’, Papa always said.

Resolves #172

### DIFF
--- a/src/scripts/scheduleJobs.js
+++ b/src/scripts/scheduleJobs.js
@@ -3,8 +3,10 @@ import logger from '../server/utils/logger'
 
 const scheduleJobs = queueDict => queueDict.scheduler.scheduleJobs()
 
+const isJob = job => (!!job && job.constructor.name === 'Job')
+
 const renderScheduledJobs = scheduledJobs => scheduledJobs
-  .filter(job => !!job)
+  .filter(isJob)
   .forEach(job => logger.info(`Scheduled job: ${job.toKey()}`))
 
 const scheduleableQueues = queueDicts.filter(queueDict => queueDict.scheduler.getIsScheduled())
--- a/src/scripts/scheduleJobs.js
+++ b/src/scripts/scheduleJobs.js
@@ -3,14 +3,15 @@ import logger from '../server/utils/logger'
 
 const scheduleJobs = queueDict => queueDict.scheduler.scheduleJobs()
 
-const renderResults = (results) => {
-  results.forEach(result => logger.info(`Scheduled job: ${result.name}`))
-}
+const renderScheduledJobs = scheduledJobs => scheduledJobs
+  .filter(job => !!job)
+  .forEach(job => logger.info(`Scheduled job: ${job.toKey()}`))
 
-const promises = queueDicts.map(scheduleJobs)
+const scheduleableQueues = queueDicts.filter(queueDict => queueDict.scheduler.getIsScheduled())
+const promises = scheduleableQueues.map(scheduleJobs)
 Promise.all(promises)
-  .then((results) => {
-    renderResults(results)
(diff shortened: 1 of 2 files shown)


src/scripts/scheduleJobs.js
```diff
@@ -3,14 +3,15 @@ import logger from '../server/utils/logger'
 
 const scheduleJobs = queueDict => queueDict.scheduler.scheduleJobs()
 
-const renderResults = (results) => {
-  results.forEach(result => logger.info(`Scheduled job: ${result.name}`))
-}
+const renderScheduledJobs = scheduledJobs => scheduledJobs
+  .filter(job => !!job)
```
slifty: I like it -- but lets put it in a named function to make it very clear what we're trying to do (it's clever, named function will explain what's up)

src/server/queues/AbstractJobScheduler.js
```diff
@@ -1,4 +1,7 @@
-import { QUEUE_SCHEDULER_TIMEZONE } from './constants'
+import {
+  Schedules,
```
slifty: whoops that should be upper case shouldn't it.  my bad (out of scope for this PR)

reefdog: Haha yup, see the bottom of this comment: https://github.com/TechAndCheck/tech-and-check-alerts/issues/150#issuecomment-522090316

src/scripts/scheduleJobs.js
reefdog: Did this, but also added an additional check for job-ishness (since `isJob` felt right but also seemed a little grand when we were really just checking for truthiness before).

PR #177, opened 2019-08-20
Extract, store, and display CNN show name

### LINKED ISSUE #128: Move scraper decorator methods to AbstractStatementScraper
We have `addCanonicalUrlToStatements` and `addScraperNameToStatements` and neither concept is specific to the cnnScraper.  We should move those decorators to the AbstractStatementScraper.

Since these are universal needs for all scraped statements, we should also have the scrapeHandler at the AbstractStatementScraper invoke them as part of the return value, rather than relying on the implementing class to remember to invoke them on the pipeline.

### LINKED ISSUE #149: Extract, store, and display CNN show name
The existing scraper [maps the show identifier in the URL](https://github.com/TechAndCheck/tech-and-check-alerts-2018/blob/ab1bb599319f034dbe300d2890b1638df764d89e/cnn_tools/cnn_buster.py#L49-L53) to a [fixed list of shows](https://github.com/TechAndCheck/tech-and-check-alerts-2018/blob/master/cnn_tools/shows.json) and displays the show name within the newsletter. We should be doing the same.

reefdog:
We were forward-thinking and added a `source` field to Claims for storing a scraper-specific value as the "source" of the claim. For CNN, the source is the show that the claim appeared on. (For Twitter, this will be the username of the account that posted the claim, though we won't actually display this on the newsletter.)

CNN displays the show name on the transcript page. We now scrape that show name, store it on the statement, and display it on the transcript:

cnn-source

A few other things that came along for the ride:
* `CnnTranscriptStatementScraper#addScraperNameToStatements()` and `CnnTranscriptStatementScraper#addCanonicalUrlToStatements()` were moved up to the `AbstractStatementScraper` and combined into a single `addSchemaPropertiesToStatements()` method, so we could do a single loop over the statements instead doing one for every additional property. (Consequently, we were able to resolve #128.)
* The getting and setting of statement `canonicalUrl` and `source` were abstracted more cleanly and to allow for statement-specific determination of those values.
* `AbstractScraper`'s setting of `this.scrapeUrl` was moved from an inline constructor action to a `setScrapeUrl()` setter method that essentially tries to enforce immutability, since a scraper's `scrapeUrl` should never change after instantiation.
* We now store the scrape response on the scraper itself as `this.scrapeResponse` so that methods that can't practically be passed the scrape response as input can still use it.
* The [`memoize-one`](https://github.com/alexreardon/memoize-one) library was added so we can memoize functions that will be called repeatedly with the same input and short-circuit processing the entire function again.

Side note: CNN also includes a show-specific shortcode in the transcript URL; extracting that and mapping that to a static list of known shows was our previous way of determining show name, but we're abandoning that for the increased responsiveness and flexibility of getting the show name from the transcript page itself.

Resolves #149
Resolves #128


src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -59,8 +64,7 @@ class CnnTranscriptStatementScraper extends AbstractStatementScraper {
       normalizeStatementSpeakers,
       removeNetworkAffiliatedStatements,
       removeUnattributableStatements,
-      this.addScraperNameToStatements,
-      this.addCanonicalUrlToStatements,
+      this.addAdditionalPropertiesToStatements,
```
slifty: While we're refactoring this... we should be doing it in the AbstractStatementScraper see #128 

Happy to clarify what I mean by slack.

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -31,6 +32,8 @@ class CnnTranscriptStatementScraper extends AbstractStatementScraper {
 
   getScraperName = () => STATEMENT_SCRAPER_NAMES.CNN_TRANSCRIPT
 
+  getSource = () => extractSourceFromTranscriptUrl(this.getScrapeUrl())
```
slifty: `getSource` should also be an abstract method defined in AbstractStatementScraper (since all statements will have a source)

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -41,11 +44,13 @@ class CnnTranscriptStatementScraper extends AbstractStatementScraper {
     return bodyTexts[2]
   }
 
-  addScraperNameToStatements = statements => statements
-    .map(statement => ({ ...statement, scraperName: this.getScraperName() }))
-
-  addCanonicalUrlToStatements = statements => statements
-    .map(statement => ({ ...statement, canonicalUrl: this.scrapeUrl }))
+  addAdditionalPropertiesToStatements = statements => statements
```
slifty: This name feels... vague / generic.  Wondering if `addUniversalPropertiestoStatements` is any better... probably not.

`addSchemaProperitiesToStatements`, or what about `tellTheOriginStoryOfTheseStatements`...

Honestly we can keep it as addAdditionalProperties if we can't think of something better (I'm not sure any of these are better), I appreciate the DRY though feel the name is a step back, but it's OK.

src/server/constants.js
```diff
@@ -10,3 +10,34 @@ export const CLAIMBUSTER_THRESHHOLD = 0.5
 export const CLAIM_PLATFORM_NAMES = {
   CNN_TRANSCRIPT: 'CNN',
 }
+
+// Keys should be lower-case to match the value we pull from the URL and store in the database
+export const CNN_SHOW_NAMES = {
```
slifty: OK so I'm about to give some maybe annoying feedback.... Show name can be extracted directly from the DOM, and I kind of think that is how it should actually be pulled.  This would make things less rigid (we are using the source name, and if it changes we don't have to change code) and also mean we aren't storing a bunch of weird 3rd party content in our pristine constants file.

Example: 
image

It's under an element with the class "cnnTransStoryHead"

What do you think...

src/server/utils/cnn.js
```diff
@@ -54,6 +58,24 @@ export const extractPublicationDateFromTranscriptUrl = (url) => {
...
+/**
+ * Extracts and returns the CNN show shortcode from CNN transcript URLs.
+ *
+ * For example, the URL `http://transcripts.cnn.com/TRANSCRIPTS/1908/13/cnr.01.html` contains the
+ * shortcode `cnr`, which identifies the show as CNN Newsroom.
+ *
+ * @param  {String} url The transcript URL to parse
+ * @return {String}     The extracted source shortcode
+ */
+export const extractSourceFromTranscriptUrl = (url) => {
```
slifty: I wouldn't call it a source here, but a shortcode -- the fact that we are using it as the source is somewhat secondary and linking it to this name makes the method less useful for other purposes.

src/server/utils/cnn.js
```diff
@@ -23,11 +23,15 @@ const attributionAffiliationRegex = /,([^a-z]*)/
...
-export const isTranscriptListUrl = url => url.startsWith('/TRANSCRIPTS/')
-  && url.endsWith('.html')
+export const isTranscriptListUrl = url => (url.startsWith('/TRANSCRIPTS/')
+  || url.startsWith('http://transcripts.cnn.com/TRANSCRIPTS/')
+  || url.startsWith('http://cnn.com/TRANSCRIPTS/')
+) && url.endsWith('.html')
 
-export const isTranscriptUrl = url => url.startsWith('/TRANSCRIPTS/')
-  && url.endsWith('.html')
+export const isTranscriptUrl = url => (url.startsWith('/TRANSCRIPTS/')
```
slifty: This doesn't actually validate that it is a genuine URL are you sure you're OK with that?

src/server/utils/cnn.js
```diff
@@ -23,11 +23,15 @@ const attributionAffiliationRegex = /,([^a-z]*)/
 const attributionNameRegex = /[^a-z,]+/
 const chunkAttributionRegex = /[^a-z]+[:]/
 
-export const isTranscriptListUrl = url => url.startsWith('/TRANSCRIPTS/')
-  && url.endsWith('.html')
+export const isTranscriptListUrl = url => (url.startsWith('/TRANSCRIPTS/')
```
slifty: This doesn't actually validate that it is a genuine URL are you sure you're OK with that?

src/server/utils/__test__/cnn.test.js
```diff
@@ -83,6 +84,21 @@ describe('utils/cnn', () => {
     })
   })
 
+  describe('extractSourceFromTranscriptUrl', () => {
+    it('Should extract the source shortcode from a transcript url', () => {
+      expect(extractSourceFromTranscriptUrl('/TRANSCRIPTS/1908/13/cnr.01.html'))
```
slifty: Lets add some tests for http://cnn.com ones as well (and the transcripts. variety)

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
reefdog: Ah, I didn't know about #128 but this makes perfect sense. I don't need clarification, though we may want to quickly pair on implementation if you already have specific thoughts (since I have given it literally zero though thus far).

src/server/constants.js
reefdog: I think that's better in every way. I didn't realize we had a stable show name; I think @markstencel or @ericaryan had mentioned that the show name isn't consistently present in the transcript.

It doesn't solve the problem of us having to maintain a list, but maybe we look for `cnnTransStoryHead` first, then fall back on the shortcode-from-URL when that's missing? In this model, when having to use the shortcode option we'd go ahead and write the longname to the `source` field during scrape time (and fall back on `NULL` if we didn't have a mapping present). This would let me get rid of that Handlebars helper, which is very nice, but would still require we keep that shortcode mapping in the system (just for backup).

Thoughts on that?

src/server/utils/cnn.js
reefdog: I'm not but saving substantive work here for #129.

src/server/utils/cnn.js
reefdog: I'm not OK with it, but saving that for #129.

src/server/utils/__test__/cnn.test.js
reefdog: lol I started to but decided to stash it here: https://github.com/TechAndCheck/tech-and-check-alerts/issues/129#issuecomment-523202538

Really want me to go ahead with that (in a more simplistic form than I had started to)?

slifty: Don't go nuts with the tests BUT I think it's good form to test the functionality you add (since you added those two forms a test to verify that it's handling those two forms correctly would be good).

reefdog: True true. Will do!

src/server/constants.js
reefdog: After some other offline convo, our strategy:

1. Start by just using `cnnTransStoryHead` and fall back to blank.
2. Since that's optional, log a warning when blank rather than throw an error.

src/server/utils/cnn.js
reefdog: Gonna stop extracting from the URL entirely (see https://github.com/TechAndCheck/tech-and-check-alerts/pull/177#discussion_r315967711).

src/server/utils/__test__/cnn.test.js
reefdog: Rewinding all this (https://github.com/TechAndCheck/tech-and-check-alerts/pull/177#discussion_r315967711) so resolving this.

reefdog:
Okay, I just committed a large and substantive change in response to @slifty's feedback. I _think_ most of it is an improvement, but there's one area I definitely expect to be controversial and possibly need changing. I'm just not sure the right way to change it. I recommend reading the [entire commit message](https://github.com/TechAndCheck/tech-and-check-alerts/pull/177/commits/61545643168930ffacfa12ef546789f28e6b0249), but the controversial bit:

> One probable area of concern is the use of
`CnnTranscriptStatementScraper.setSource()` and `this.source`, which is
not modeled on `AbstractStatementScraper`. I'm hitting an OOP design
wall and could use some input. I'm leaning toward modeling a Statement
object generally, applying that schema (as properties only, with null
values) to `this.schema` in the `AbstractStatementScraper` constructor,
and then letting `set{Property}()` setters modify that object. Feedback
welcome!

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -31,6 +32,23 @@ class CnnTranscriptStatementScraper extends AbstractStatementScraper {
 
   getScraperName = () => STATEMENT_SCRAPER_NAMES.CNN_TRANSCRIPT
 
+  getCanonicalUrl = () => this.getScrapeUrl()
+
+  getSource = () => this.source
+
+  setSource = (source) => { this.source = source }
```
slifty: I think it would be cleaner to remove setSource and instead add a cache to getShowName (or... we could use [memoize-one](https://www.npmjs.com/package/memoize-one) and skip writing our own caching logic.)

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -2,12 +2,38 @@ import AbstractScraper from './AbstractScraper'
...
-   * scraped page.
+   * The source is the specific context in which the statement appeared, as opposed to the platform
+   * it was delivered via or the speaker who delivered it. For example, on TV, this would be a show
+   * title; on social media, it would be the username of the poster.
+   *
+   * OVERRIDE WHEN EXTENDING
+   *
+   * @return {String} The source for the statement
+   */
+  getSource = () => {
```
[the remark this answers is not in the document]
reefdog: Will do! In both cases, for CNN specifically the statement will be ignored and the values determined by class attributes (`this.scrapeUrl` and whatever we call the `responseString` one… `this.scrapeResponse` or whatever… both via getter methods of course), but I grok that for other scrapers (like Twitter) the statement _will_ contain the values necessary, so I'll model it as such.

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -17,7 +43,27 @@ class AbstractStatementScraper extends AbstractScraper {
...
+  addSchemaPropertiesToStatements = statements => statements
+    .map(statement => ({
+      ...statement,
+      scraperName: this.getScraperName(),
+      canonicalUrl: this.getCanonicalUrl(),
+      source: this.getSource(),
+    }))
+
+
+  scrapeHandler = (responseString) => {
```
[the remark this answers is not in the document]
reefdog: Is `scrapeResponse` an acceptable attribute name? I'll add accessor methods for it of course.

And yes, will document the immutability of `scrapeUrl`.

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -41,12 +59,6 @@ class CnnTranscriptStatementScraper extends AbstractStatementScraper {
     return bodyTexts[2]
   }
 
-  addScraperNameToStatements = statements => statements
```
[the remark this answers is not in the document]
reefdog: You stay gone since you moved up to an abstract class.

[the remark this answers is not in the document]
reefdog: Boom roasted.

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
reefdog: Perfect use of memoize-one.

reefdog:
Okay, feedback has been implemented, ready for re-review! This is a pretty cool set of changes:

1. `scrapeUrl` is now documented as and attempts to be treated as immutable.
2. The scrape response is stored on the scraper as `scrapeResponse`, to be available to methods that need it but can't practically be passed it.
3. `getSource()`/`getCanonicalUrl()` have had their abstract footprint modified to accept a `statement` (since those properties will often be determined by the statement itself, rather than shared across the bulk statement scraper the way it is with CNN), and renamed to `getStatement{Property}()`
4. Replace the bad idea of setting `this.source` on `CnnTranscriptStatementScraper` with the good idea of a memoized version of the `getShowName()` function.

I messed up the commit title on e36a1e3df6a6345bc36f44fe58a045e4acd3f96b, by which I mean I forgot to give it one, but we lucked into the first line of the description truncating in a sensible way.

One thing we specifically _didn't_ do here is replace all the times we're passing around `responseString` as a parameter with locally accessing `this.getScrapeResponse()` w/in those methods. @slifty brings up the point that it may be an antipattern to mask a method's reliance on vital external data. Punting that 🤔 for later.

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -17,7 +49,27 @@ class AbstractStatementScraper extends AbstractScraper {
...
+   * and apply all these additional properties.
+   *
+   * @param {Object[]} statements The statements that have been scraped
+   * @return {Object[]}           Those same statements with additional schema properties added
+   */
+  addSchemaPropertiesToStatements = statements => statements
+    .map(statement => ({
+      ...statement,
+      scraperName: this.getScraperName(),
+      canonicalUrl: this.getStatementCanonicalUrl(),
```
slifty: Don't forget to pass in the statement!

Also we may want to consider only calling getStatementCanonicalUrl if the statement canonicalUrl isn't already set.

So

`canonicalUrl: statement.canonicalUrl | this.getStatementCanonicalUrl(statement)`

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -17,7 +49,27 @@ class AbstractStatementScraper extends AbstractScraper {
...
+      source: this.getStatementSource(),
```
slifty: Same as above

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -34,6 +37,37 @@ class CnnTranscriptStatementScraper extends AbstractStatementScraper {
...
+   * Currently, CNN transcripts are organized by show, so all statements from a single transcript
+   * scrape share the same source.
+   *
+   * Rather than rerun `getShowName()` for every statement from the transcript, we re-use the first
+   * calculated value by wrapping it in a memoizer.
+   *
+   * @param {String} html The scraped transcript HTML
+   * @return {String}     The show name or an empty fallback string
+   */
+  memoizedGetShowName = memoizeOne(this.getShowName)
```
slifty: Memoizing is a bit of an internal feature so not part of the API -- I think we should actually just have

```
getShowName = memoizeOne((html) => {...})
```

reefdog:
Ready for re-re-re-review!

PR #179, opened 2019-08-20
Remove `REP.` prefix from speaker name

### LINKED ISSUE #178: `REP.` doesn't get removed from speaker name
From a test email just now:

> REP. STEVE COHEN on CNN Newsroom (CNN):
> We did it with Obama, so I'd take it into consideration but I'd be concerned about it because it would increase the debt, which we -- I voted against the Trump tax bill, which put over a trillion dollars on the debt, and I voted against other proposals he's had, the budget proposal, put over a trillion dollars on the debt.

This is an easy one. We [strip out `REPRESENTATIVE`](https://github.com/TechAndCheck/tech-and-check-alerts/blob/fd414be2d9ad5b26f4d8e88d7f290f0ea03f176b/src/server/utils/cnn.js#L216-L218) but not `REP. `; need to do so.

reefdog:
Straightforward: we strip `REPRESENTATIVE`, `SENATOR`, and `SEN.` but missed `REP.`.

Resolves #178

### DIFF
--- a/src/server/utils/__test__/cnn.test.js
+++ b/src/server/utils/__test__/cnn.test.js
@@ -372,6 +372,8 @@ describe('utils/cnn', () => {
         .toBe('JIMMY DUST')
       expect(cleanSpeakerName('REPRESENTATIVE JIMMY DUST'))
         .toBe('JIMMY DUST')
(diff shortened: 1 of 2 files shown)


reefdog:
Stripped of all honorifics.

PR #182, opened 2019-08-21
Collapse repeated whitespace to a single space

### LINKED ISSUE #106: Squish whitespace within claims
Because of line breaks and such, claims often end up with consecutive whitespace:

```
And I'm terribly worried that telling people, 180 million Americans, that they will no longer have the health insurance that they've earned at their   job, "But we're going to give you something better" -- actually believe it'd be better -- but that's a big leap of faith for 180 million Americans   to take.
```

We should squish and trim the whitespace.

reefdog:
Because of line breaks and formatting, we often end up with claims full of multiple spaces between words. This gets visually collapsed to a single space in most rendering situations (like HTML), but we shouldn’t rely on that, and we should clean these down to single spaces when possible.

This adds a generic `squish()` utility (named after Ruby’s), plus a scraper utility for iterating over statements and squishing their text, plus tests, and then adds that squish utility to the `CnnTranscriptStatementScraper` sequence of statement-cleaning functions.

Resolves #106

### DIFF
--- a/src/server/utils/__test__/index.test.js
+++ b/src/server/utils/__test__/index.test.js
@@ -1,5 +1,8 @@
 import {
-  isTestEnv, isDevelopmentEnv, isProductionEnv,
+  isTestEnv,
+  isDevelopmentEnv,
+  isProductionEnv,
+  squish,
 } from '..'
 
 describe('utils/index', () => {
@@ -16,6 +19,11 @@ describe('utils/index', () => {
   describe('#isProductionEnv', () => {
     it('Should report that we are in not in the production environment', () => {
       expect(isProductionEnv()).toBe(false)
+    })
+  })
+  describe('#squish', () => {
+    it('Should replace all whitespace with single spaces', () => {
+      expect(squish('Too  much   space.\nAmiright?')).toBe('Too much space. Amiright?')
     })
   })
 })
--- a/src/server/utils/__test__/scraper.test.js
+++ b/src/server/utils/__test__/scraper.test.js
@@ -1,6 +1,7 @@
 import dayjs from 'dayjs'
 import {
   isDateBeyondScrapeHorizon,
+  squishStatementsText,
 } from '../scraper'
 
 import config from '../../config'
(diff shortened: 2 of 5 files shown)


src/server/utils/__test__/index.test.js
```diff
@@ -18,4 +21,9 @@ describe('utils/index', () => {
       expect(isProductionEnv()).toBe(false)
     })
   })
+  describe('#squish', () => {
+    it('Should replace all whitespace with single spaces', () => {
+      expect(squish('Too  much   space.\nAmiright?')).toBe('Too much space. Amiright?')
```
slifty: uh, this directly offends me since, as we all know, you are supposed to double space after periods.

In srsns is it expected behavior to convert newlines to spaces?

src/server/utils/__test__/scraper.test.js
```diff
@@ -22,3 +23,17 @@ describe('isDateBeyondScrapeHorizon', () => {
       .toBe(false)
   })
 })
+
+describe('squishStatementsText', () => {
```
slifty: generals attorney is dumb and this made me angry because it made me think of that.

It's right, just wanted to complain.

src/server/utils/index.js
```diff
@@ -17,3 +17,14 @@ export const runPromiseSequence = promiseArray => promiseArray.reduce(
 )
 
 export const runSequence = (sequence, seed) => sequence.reduce((param, fn) => fn(param), seed)
+
+/**
+ * Squish a string by collapsing repeated whitespace to a single space.
+ *
+ * This has the practical effect and primary purpose of reducing multiple spaces to single spaces,
+ * but as a side effect it will replace tabs and newlines with a space.
```
slifty: OK you address this -- It looks like this is the expected behavior in Rails too.  One thought: listing this as a "side effect" feels odd because... isn't that the intended effect (side effect makes it feel like a bug)?  (this is way too nit picky even for me so just pretend I said nothing.)

src/server/utils/__test__/index.test.js
reefdog: I know you jpf but for real I did feel bad that it essentially imposes a single-space-after-period preference. (Which is to say, squishing paragraphs isn't the intent; squishing sentences is.) But yes, as you mention later, this is `squish()` behaving properly; the real question is whether we want to trouble ourselves with squishing the right thing in the first place.

src/server/utils/index.js
reefdog: Yeah, good point. I incorrectly mixed documentation of "how we intend to _use_ `squish()`" (where the tab/newline collapsing is a side effect, not the point) into the generic documentation of the function itself. I'll remove the side effect language!

src/server/utils/__test__/scraper.test.js
reefdog: It's one of my favorite right wrong things.

PR #203, opened 2019-09-01
Add twitter scraper queue

slifty:
This creates a queue for scraping twitter accounts and ultimately converting those scrapes into claims.  This queue is NOT an entry point for the twitter scrape process (e.g. another process is going to have to create jobs that specify which accounts to scrape), and as such it is not a scheduled process.

This PR introduces a new model, the `statement_log`, which allows the system to keep track of statements that had been processed in the past.  As such, we need to run `yarn migrate` after merging this PR.

Some quick notes about statement logs:

1.  They do not actually store the statement, but rather store the canonical URL of the statement.
2. We may want to allow the developer to bypass logging (this also applies to scraper logs, frankly).  Currently if the developer invokes a statement scrape for testing purposes the resulting statements will be marked as scraped / processed in the entire system even though technically they did not get processed in full.


slifty:
Oh good lord I accidentally all the commits in this branch.

Well, enjoy reviewing all of this lolololol (just kidding I'll try to cherry pick the later commits into the separate branch as intended)

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -64,11 +87,29 @@ class AbstractStatementScraper extends AbstractScraper {
...
+  scrapeHandler = async (responseString) => {
+    const decorationSequence = [
+      this.addSchemaPropertiesToStatements,
+      this.addMetaPropertiesToStatements,
+    ]
     const statements = this.statementScrapeHandler(responseString)
-    const statementsWithCompletedSchemas = this.addSchemaPropertiesToStatements(statements)
-    return statementsWithCompletedSchemas
+    const decoratedStatements = await runAsyncSequence(decorationSequence, statements)
+    this.createStatementLogs(decoratedStatements)
```
slifty: Note: this will create a new statement log EVERY TIME -- even if the statement had been seen many times before.

This is the price we pay for having the scraper do the logging (rather than the pipeline).  It does mean the log table is going to be full of a lot of junk.  We may want to change the log behavior to only insert when the item hadn't been seen before.

src/server/models/StatementLog.js
```diff
@@ -0,0 +1,18 @@
+module.exports = (sequelize, DataTypes) => {
+  const StatementLog = sequelize.define('StatementLog', {
+    canonicalUrl: DataTypes.STRING(1024),
+    scraperName: DataTypes.STRING,
+  }, {})
+
+  StatementLog.getMostRecentStatementLogByStatement = async statement => (
```
reefdog: Ooh, I like this. We should do the same (separate cleanup issue) for `ScrapeLog`, moving some of `AbstractScraper`'s `getMostRecentScrapeTime()` and `getMostRecentSuccessfulScrapeTime()` to `ScrapeLog` class methods.

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -49,6 +53,25 @@ class AbstractStatementScraper extends AbstractScraper {
     throw new Error('You implemented a statement scraper but forgot to define the statementScrapeHandler.')
   }
 
+  /**
+   * Generates and saves a statement log entry for each distinct canonical url.
+   * This is used to keep track of which statements have been seen by the system
+   * (since not all statements end up being claims).
+   *
+   * @param  {Array} The array of statements to log
```
reefdog: Need to provide the param name (`statements`) here for valid JSDoc. Right now it thinks `The` is the param name. Also you document it as `Array` but we tend to do `Object[]` for statements, yeah?

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -49,6 +53,25 @@ class AbstractStatementScraper extends AbstractScraper {
...
+   * @return {StatementLog} the saved initial statement log entries
```
reefdog: I wasn't going to mention this but since you're touching the line above anyway: `the` → `The`.

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -49,6 +53,25 @@ class AbstractStatementScraper extends AbstractScraper {
...
+   */
+  createStatementLogs = (statements) => {
+    const uniqueUrlSet = new Set(
+      statements.map(statement => statement.canonicalUrl),
+    )
+    return [...uniqueUrlSet]
+      .map(canonicalUrl => StatementLog.create({
+        canonicalUrl,
+        scraperName: this.getScraperName(),
+      }))
```
reefdog: Can we save some cycles/connections here by using [`StatementLog.bulkCreate([...uniqueUrlSet]]`](https://sequelize.org/master/manual/instances.html#working-in-bulk--creating--updating-and-destroying-multiple-rows-at-once-)?

Note that if we did so, this should return the promise (so document accordingly), and then `scrapeHandler()` could choose if/how to handle the promise (at least catching any errors).

src/server/models/StatementLog.js
reefdog: Reiterating that this is a kudos comment and observation, not a task. I'll even go ahead and mark it as resolved.

src/server/workers/scrapers/AbstractStatementScraper.js
slifty: Good call -- I'm updating the method to be async (which means I think I can leave the documentation the same?  how do you document return values of async methods...)

slifty:
I realized a good way to avoid noisy logging is to take advantage of something that already exists -- filter the list of statements by the existance of mostRecentLog before ordering the new log lines.

I'll also change that name since we will only store a single log per statement, so it would be something like "isNovel" or "hasBeenLogged" and make it a bool rather than a full blown StatementLog entry.

src/server/workers/scrapers/AbstractStatementScraper.js
reefdog: [Officially, you can just `@returns {Promise}`.](https://jsdoc.app/tags-returns.html) But it looks like `@returns {Promise<Array>}` is currently supported, with a more explicit syntax (that also lets you document specific error thrown) coming. See https://github.com/jsdoc/jsdoc/issues/1197#issuecomment-312948746.

src/server/workers/scrapers/AbstractStatementScraper.js
slifty: This is no longer the case; so resolving.

src/server/workers/scrapers/AbstractStatementScraper.js
```diff
@@ -49,6 +54,28 @@ class AbstractStatementScraper extends AbstractScraper {
...
+    const filteredStatements = filterPreviouslyScrapedStatements(statements)
+    const uniqueUrlSet = new Set(
+      filteredStatements.map(statement => statement.canonicalUrl),
+    )
+    return StatementLog.bulkCreate(
+      [...uniqueUrlSet]
+        .map(canonicalUrl => ({
+          canonicalUrl,
+          scraperName: this.getScraperName(),
+        })),
```
reefdog: I'm a little uncomfortable with doing this much transformation inline as the parameter to `bulkCreate()`. I like the conciseness, but it does require a bit of nested reading. Would you consider:
```js
const decoratedStatements = [...uniqueUrlSet]
  .map(canonicalUrl => ({
    canonicalUrl,
    scraperName: this.getScraperName(),
  }))
return StatementLog.bulkCreate(decoratedStatements)
```

slifty: Fixed up!  I also went ahead broke out the url extraction to a helper method.

PR #210, opened 2019-09-03
Add support for ClaimBuster API V2

### LINKED ISSUE #139: Use ClaimBuster v2 API
The ClaimBuster team has a new version of the API and would like us to start using it, preferably by the end of the month. See https://idir.uta.edu/claimbuster-dev/api/ and https://idir.uta.edu/claimbuster-dev/api/docs/. Our API key is in the vault.

We should keep v1 API support available for a while and make API version env-configurable.

reefdog:
This adds optional support for V2 of the ClaimBuster API, which is enabled by setting the appropriate `.env` variable.

The new API has a different URL structure and requires an API key, so this also adds two `ClaimBusterClaimDetector` instance methods to help compose the response and smooth out the differences. (I struggled with whether these should be utility functions, but decided in the event we wanted to do things like abstract out ClaimBuster integration into a base class for calling other endpoints, it belonged here.)

This also moves the `CLAIMBUSTER_THRESHHOLD` from the shared constants file into a `claimDetector`-specific one. This may be a move in the wrong direction, but at least it's consistent with our other compartmentalization of constants.

Note that after merging this, you'll want to add the appropriate new `.env` variables (noted in `.env.template`), or else it will just default to V1.

Resolves #139


reefdog:
Fair! My working assumption was that we'd want an eject button to push if we suddenly discovered Big Problems with v2. But doing a clean conversion is substantially simpler, and this is a good time to do it anyway. (And recovering from it won't be _that_ hard.) I'll remove the v1 split and make it clean!

reefdog:
Not really worth muddying commit history with the direction taken in this PR, so just gonna close it and delete the branch. #217 foreverrrrrr!

Issue #47, opened 2019-05-30
Populate TESTING.md


slifty:
We should quickly document how to add new tests to a docs/TESTING.md -- this is linked to from our [CONTRIBUTING.md](https://github.com/TechAndCheck/tech-and-check-alerts/blob/master/CONTRIBUTING.md) already but 404s.

It should provide really simple instructions on the way to properly introduce a new test to a new component.

PR #220, opened 2019-09-05
Automated automated testing for CNN pipeline

### LINKED ISSUE #127: Adjust cnn scraper test suite to use test data
As part of #125 we will probably need / want to adjust the way the unit tests are written as well, and Instead of having a long list of manually populated we should create test data which represents the "state" at different points of the pipeline.

This way our tests of each pipeline utility method can iterate through the data that has the correct format for each step.

This also will fix the situation where discovering a new edge case would require a whole bunch of duplicate test code across different (similar) components (e.g. "getNameFromAttribution" and "getAffiliationFromAttribution")

slifty:
This PR creates a mechanism (and utility) for us to easily expand the CNN test suite to include real world data as we find instances of the system failing.

For documentation on how to use these tools meaningfully go check the new `docs/TESTING.md` which is part of this PR.

run `yarn generate:test:cnn -u [FULL_TRANSCRIPT_URL_HERE]`

Resolves #127 
Related to #47 since we technically now have TESTING.md

### DIFF
--- a/src/scripts/generateCnnUtilTest.js
+++ b/src/scripts/generateCnnUtilTest.js
@@ -1,0 +1,157 @@
+import fs from 'fs'
+import GenericScraper from '../server/workers/scrapers/GenericScraper'
+import {
+  isTranscriptUrl,
+  getTranscriptTextFromHtml,
+  removeTimestamps,
+  removeSpeakerReminders,
+  removeDescriptors,
+  addBreaksOnSpeakerChange,
+  splitTranscriptIntoChunks,
+  extractStatementsFromChunks,
+  cleanStatementSpeakerNames,
+  normalizeStatementSpeakers,
+  removeNetworkAffiliatedStatements,
+  removeUnattributableStatements,
+} from '../server/utils/cnn'
+import logger from '../server/utils/logger'
+
+const FILE_ENCODING = 'utf8'
+const BASE_PATH = `${__dirname}/../server/utils/__test__/data/cnn`
+
+const scrapeSequence = [
+  getTranscriptTextFromHtml,
+  removeTimestamps,
+  removeSpeakerReminders,
+  removeDescriptors,
+  addBreaksOnSpeakerChange,
+  splitTranscriptIntoChunks,
+  extractStatementsFromChunks,
+  cleanStatementSpeakerNames,
+  normalizeStatementSpeakers,
+  removeNetworkAffiliatedStatements,
+  removeUnattributableStatements,
+]
+
+const getTestSuiteIndex = () => {
+  const indexFlag = process.argv.indexOf('-i')
+  if (indexFlag !== -1) {
+    const suiteIndex = process.argv[indexFlag + 1]
+    return suiteIndex
+  }
+  return null
+}
+
+const getUrl = () => {
+  const urlFlag = process.argv.indexOf('-u')
+  if (urlFlag !== -1) {
+    const url = process.argv[urlFlag + 1]
+    if (isTranscriptUrl(url)) {
+      return url
+    }
+  }
+  return null
+}
+
+const getTestSuiteDirectoryPath = suiteIndex => `${BASE_PATH}/${suiteIndex}`
+
+const getTestSuiteUrlPath = (suiteIndex) => {
+  const testSuiteDirectoryPath = getTestSuiteDirectoryPath(suiteIndex)
+  return `${testSuiteDirectoryPath}/_url.txt`
+}
+
+const getTestSuiteStepPath = (suiteIndex, stepIndex, stepName) => `${getTestSuiteDirectoryPath(suiteIndex)}/${stepIndex}_${stepName}.json`
+
+const getUrlFromTestSuite = (suiteIndex) => {
+  try {
+    return fs.readFileSync(getTestSuiteUrlPath(suiteIndex), FILE_ENCODING)
+  } catch {
+    return null
+  }
+}
+
+// This method is a bit magic.  It gets a list of all existing files in the suites directory
+// and then finds the file with the highest number name (index).
+const getNextTestSuiteIndex = () => {
+  const testSuiteIndexes = fs
+    .readdirSync(BASE_PATH) // Get the files
+    .map(fileName => parseInt(fileName, 10)) // Convert the names to numbers
+    .filter(Number.isInteger) // Get rid of anything that wasn't a number
+    .sort((a, b) => a - b) // Sort numerically
+  return testSuiteIndexes.length === 0 ? 1 : testSuiteIndexes.slice(-1)[0] + 1
+}
+
+const createTestSuiteDirectory = (suiteIndex) => {
+  const dirPath = getTestSuiteDirectoryPath(suiteIndex)
+  try {
+    if (!fs.existsSync(dirPath)) {
+      fs.mkdirSync(dirPath)
+    }
+  } catch (e) {
+    logger.error(e)
+  }
+}
+
+const saveTestUrl = (suiteIndex, url) => {
+  try {
+    fs.writeFileSync(
+      getTestSuiteUrlPath(suiteIndex),
+      url,
+      { encoding: FILE_ENCODING },
+    )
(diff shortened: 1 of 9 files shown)


src/scripts/generateCnnUtilTest.js
```diff
@@ -0,0 +1,145 @@
...
+  normalizeStatementSpeakers,
+  removeNetworkAffiliatedStatements,
+  removeUnattributableStatements,
+} from '../server/utils/cnn'
+import logger from '../server/utils/logger'
+
+const FILE_ENCODING = 'utf8'
+const BASE_PATH = `${__dirname}/../server/utils/__test__/data/cnn`
+
+const scrapeSequence = [
```
slifty: @reefdog what do you think about me updating the CnnStatementScraper to have its sequence actually be a class variable, so this could be scrapeSequence = CnnStatementScraper.scrapeSequence

That way if the sequence changes we don't need to update that change in three places...

src/scripts/generateCnnUtilTest.js
```diff
@@ -0,0 +1,145 @@
...
+
+const createTestSuiteDirectory = (suiteIndex) => {
+  const dirPath = getTestSuiteDirectoryPath(suiteIndex)
+  if (!fs.existsSync(dirPath)) {
+    fs.mkdirSync(dirPath)
+  }
+}
+
+const saveTestUrl = (suiteIndex, url) => {
+  fs.writeFileSync(
```
reefdog: Worth wrapping in a try/catch.

src/scripts/generateCnnUtilTest.js
```diff
@@ -0,0 +1,145 @@
...
+    .map(fileName => parseInt(fileName, 10))
+    .filter(Number.isInteger)
+    .sort((a, b) => a - b)
+  return testSuiteIndexes.length === 0 ? 1 : testSuiteIndexes.slice(-1)[0] + 1
+}
+
+const createTestSuiteDirectory = (suiteIndex) => {
+  const dirPath = getTestSuiteDirectoryPath(suiteIndex)
+  if (!fs.existsSync(dirPath)) {
+    fs.mkdirSync(dirPath)
```
reefdog: Worth wrapping in a try/catch. (All filesystem changes should be IMO.)

src/scripts/generateCnnUtilTest.js
```diff
@@ -0,0 +1,145 @@
...
+  fs.writeFileSync(
+    getTestSuiteUrlPath(suiteIndex),
+    url,
+    { encoding: FILE_ENCODING },
+  )
+}
+
+const saveTestStep = (suiteIndex, stepIndex, name, input, output) => {
+  const testStepContent = {
+    name,
```
reefdog: It's recommended we not use `name`; it's not technically a reserved word, but it is on the list of common properties to avoid using as constant names.

src/server/utils/__test__/data/cnn/index.js
```diff
@@ -0,0 +1,41 @@
...
+const testSuiteBasePath = `${__dirname}`
+const testSuites = {}
+
+const getFunctionNameFromFileName = fileName => path
+  .parse(fileName).name
+  .split('_')[1]
+
+const loadTestSuite = suitePath => fs
+  .readdirSync(suitePath)
+  .filter(fileName => fileName.endsWith('.json'))
```
reefdog: Wanna move this to an `isJsonFile()` utility function (in this file)?

src/scripts/generateCnnUtilTest.js
```diff
@@ -0,0 +1,145 @@
...
+    return null
+  }
+}
+
+const getNextTestSuiteIndex = () => {
+  const testSuiteIndexes = fs
+    .readdirSync(BASE_PATH)
+    .slice(0, -1)
+    .map(fileName => parseInt(fileName, 10))
+    .filter(Number.isInteger)
```
reefdog: It's not quite clear to me why, if you're specifically parsing them as integers one line up, you'd need to filter them to only integers. Is that to get rid of `NaN`s or other weirdness that slip through?

src/scripts/generateCnnUtilTest.js
```diff
@@ -0,0 +1,145 @@
...
+
+const getUrlFromTestSuite = (suiteIndex) => {
+  try {
+    return fs.readFileSync(getTestSuiteUrlPath(suiteIndex), FILE_ENCODING)
+  } catch {
+    return null
+  }
+}
+
+const getNextTestSuiteIndex = () => {
```
reefdog: The body of this function is not super legible, but I also don't think you should expand the code too much to be self-documenting. Can you just add a note about how this function does what it does? The other functions in this file are more legible and straightforward, this one feels a bit magical (despite the obvious name).

src/scripts/generateCnnUtilTest.js
```diff
@@ -0,0 +1,145 @@
...
+const scrapeRawTranscript = (url) => {
+  const scraper = new GenericScraper(url)
+  return scraper.run()
+}
+
+const testSuiteIndex = getTestSuiteIndex()
+const currentIndex = testSuiteIndex || getNextTestSuiteIndex()
+const currentUrl = getUrl() || getUrlFromTestSuite(currentIndex)
+
+if (!currentUrl) {
```
reefdog: We have `isTranscriptUrl()`; we should use that here to avoid head-scratching situations like the one in the demo.

src/scripts/generateCnnUtilTest.js
```diff
@@ -0,0 +1,145 @@
...
+  logger.error('Pass in a valid CNN transcript URL with the `-u` parameter, or a suite index to re-run with the `-i` parameter')
+  process.exit()
+}
+
+createTestSuiteDirectory(currentIndex)
+saveTestUrl(currentIndex, currentUrl)
+scrapeRawTranscript(currentUrl)
+  .catch(error => logger.error(`Scraping failed. ${error}`))
+  .then(scrapeResult => generateTestSuiteFromSequence(currentIndex, scrapeSequence, scrapeResult))
+  .finally(() => process.exit())
```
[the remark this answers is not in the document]
slifty: Yes.

src/server/workers/scrapers/CnnTranscriptStatementScraper.js
```diff
@@ -97,7 +88,7 @@ class CnnTranscriptStatementScraper extends AbstractStatementScraper {
   }
 
   statementScrapeHandler = (responseString) => {
-    const transcript = this.getTranscriptText(responseString)
+    const transcript = getTranscriptTextFromHtml(responseString)
```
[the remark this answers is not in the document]
slifty: Yes indeed!  I'll go forward with that

src/scripts/generateCnnUtilTest.js
slifty: YOU'RE not technically a reserved word

src/scripts/generateCnnUtilTest.js
slifty: there were NaN's -- .DS_Store, _url.txt, and in theory if there was ever a .json file with a string character it would break stuff.  This is defensive programming at this point since the main culprits are filtered earlier I believe.

src/scripts/generateCnnUtilTest.js
reefdog: Good point, change it to `Justin`.

src/scripts/generateCnnUtilTest.js
```diff
@@ -0,0 +1,157 @@
...
+      { encoding: FILE_ENCODING },
+    )
+  } catch (e) {
+    logger.error(e)
+  }
+}
+
+const saveTestStep = (suiteIndex, stepIndex, stepName, input, output) => {
+  const testStepContent = {
+    name: stepName,
```
reefdog: So 😬 I think this is still a best practices violation, to have a property we made up be named `name`. This could be called anything, right? Can we call it even `fnName` or something? (Or keep it `stepName`?)

PR #222, opened 2019-09-05
Add new honorifics to filter, and improve filtering

### LINKED ISSUE #186: Filter honorifics "Governor/Gov." and "Mayor"
From today's newsletter: `MAYOR BILL DE BLASIO` and `GOV. STEVE BULLOCK`.

reefdog:
Previously, we removed honorifics from the speaker name using a form of `replace()` that didn’t ensure the honorific appeared on its own (either rooted at the beginning of the string and followed by a space, or surrounded on both sides by a space). This made it unusable for honorifics like “MAYOR” that could likely appear within someone’s name (like “SONIA SOTOMAYOR”).

This both improves upon that and lists the honorifics in an easily-appendable array, as well as adds trickier tests around them.

Resolves #186


reefdog:
That's a good idea, though I do worry about tripping over substantive initialized names, such as significant CNN contributor `J. EDGAR HOOVER`.

Issue #94, opened 2019-07-26
Add North Carolina newsletter


reefdog:
Add the newsletter for the NC-only Twitter list.

reefdog:
@markstencel @ericaryan Is the boilerplate identical for the NC newsletter? If not, could you let me know the modified text?

markstencel:
Will do. Bill is looking at the text for the national alerts (he also has suggestions for subject lines). I'll get you an NC Alert intro as soon as I get his feedback.

reefdog:
@markstencel Did you ever hear back about the boilerplate text for the NC newsletter?

PR #224, opened 2019-09-06
Add North Carolina newsletter

reefdog:
This adds the North Carolina newsletter, scheduler, and queue.

Assuming you've added the North Carolina Twitter accounts to your database (with the list name `northCarolina`) and scraped some claims, you can test this with `yarn newsletter:send-test --nc`.

Now that we have National and North Carolina in the bag, we could examine them and consider how to abstract common newsletter components even better, but that doesn’t need to be part of the scope of this.

Delivers #94


src/scripts/sendTestNewsletter.js
```diff
@@ -12,6 +13,10 @@ switch (process.argv[2]) {
   case 'national':
     send = (new NationalNewsletter()).send()
     break
+  case '--nc':
+  case 'nc':
```
slifty: I appreciate the brevity but this param is constructed differently from the national one.  what about having the options be:

--national
-n
--northcarolina
-nc

or just having it be national and northcarolina?

src/server/newsletters/NorthCarolinaNewsletter.js
```diff
@@ -0,0 +1,71 @@
...
+   *
+   * @param {Object} where The where conditions for this specific claim query
+   * @return {Object}      Query params ready for fetching claims
+   */
+  generateQueryParams = where => ({
+    where: {
+      ...where,
+      createdAt: {
+        [Sequelize.Op.gte]: dayjs().startOf('hour').subtract(1, 'day').format(),
+        // [Sequelize.Op.lt]: dayjs().startOf('hour').format(),
```
slifty: This is a rogue commented line I think (does it exist in the other template too?)

reefdog: Oh GOOD CATCH, this was _not_ meant to be committed! `SHAME`

This was me loosening the end point so I could scrape and immediately send what I just scraped. :pray:

src/scripts/sendTestNewsletter.js
reefdog: Also a good catch, and good idea on the de-brevitizing.

Since these are basically keywords identifying the thing you want (actual arbitrary parameter values), I'll keep them as hyphen-prepended parameters.

reefdog: I'm having a bad day of comprehension. Your suggested list of four (in long and short variants, all hyphen-prepended) is correct and my hyphen-less keywords are wrong. Will change…

reefdog: Wait. 🤔 In normal Unix script-calling, `-nc` would be shorthand for indicating both `-n` and `-c` options, I think? I'm not sure two-letter shorthand options are a thing, and even though this _works_ (tested to confirm), I'd hate to act in a weird way.

So I'm going to nix the shorthand, go with only longform (`--`), and ditch the hyphen-less versions entirely.

slifty: oh right that's it -- yeah lets do that one!

src/server/queues/northCarolinaNewsletterDeliveryQueue/NorthCarolinaNewsletterDeliveryQueueFactory.js
```diff
@@ -0,0 +1,10 @@
+import { QueueNames } from '../constants'
+import AbstractQueueFactory from '../AbstractQueueFactory'
+
+class NorthCarolinaNewsletterDeliveryQueueFactory extends AbstractQueueFactory {
+  getQueueName = () => QueueNames.newsletterDeliveryQueues.NATIONAL
```
slifty: This should actually be NORTH_CAROLINA (and you need a corresponding constant)

reefdog: Good caught!

Issue #167, opened 2019-08-20
Investigate error in grabbing first name of speaker attribution


ericaryan:
From Monday's test email:

S. RAMPELL (CNN):
And what's interesting is that If Trump actually thinks that there is a conspiracy happening to fake these numbers or to, you know, to fake forecasts, which a number of reporters, including Maggie Haberman and colleagues of mine at "The Washington Post" have reported, you know, the real question is, does he really think that nine other country statisticians are in on the conspiracy too because nine other countries are also -- are either in recession or on the verge of recession, and that's one of the major risks here.

http://www.cnn.com/TRANSCRIPTS/1908/19/nday.02.html

S. RAMPELL should be CATHERINE RAMPELL

Possibly caused by this?

BERMAN: So job numbers were -- you know, economic numbers were phony in the past. Now they're real. Now they're going back to being phony. You could see it in Sean's face. He knows it's B.**S. 

RAMPELL**: Yes, of course he does.

reefdog:
From today's test email:

> [ACOSTA (CNN)](http://cnn.com/TRANSCRIPTS/1908/20/cnr.20.html):
> The president is out with a new unproven conspiracy theory accusing Google of manipulating more than two million votes in 2016, adding, his victory was even bigger than thought.

The line preceding it in the transcript is `(END VIDEO CLIP)`, but also, the original instance of Jim Acosta was like this: `JIM ACOSTA, CNN CHIEF White House CORRESPONDENT`. I don't think we account for lowercase letters in the speaker.

ericaryan:
This one was from Friday:

> MEYER (CNN):
> The two biggest reasons we decided to make this move, that really troubled me, one is that every year since I've been a restauranteur in a tipping situation, the gap between what a tipped employee and someone who is legally not eligible to accept tips, also known primarily as cooks, the gap has grown and grown and grown, probably by 300 percent.

http://www.cnn.com/TRANSCRIPTS/1908/14/ampr.01.html

Original instance was DANNY MEYER, FOUNDER & CEO, UNION SQUARE HOSPITALITY GROUP

Seems like there were some weird line breaks happening in the transcript that might have played a role? Not sure.

reefdog:
Newsletter:

> [ENTEN (CNN)](http://transcripts.cnn.com/TRANSCRIPTS/1908/20/ath.01.html):
[...]

PR #231, opened 2019-09-10
Revise chunk break RegEx

slifty:
The CNN parsing pipeline uses a series of regular expressions to do
things like break a long transcript string into isolated chunks.  We
were seeing some attribution errors related to a mistake in the way
those chunks were identified.

This adjusts the RegEx to be hopefully more universal, handling the
reality that CNN sometimes includes lowercase words and letters in the
attributuon section of a statement chunk.

This also begins to move the regex into smaller re-usable components.

Issue #167
Issue #64


src/server/utils/cnn.js
```diff
@@ -21,9 +21,19 @@ const $ = cheerio
...
+const attributionAffiliationRegex = /,([^:]*)/
 const attributionNameRegex = /[^a-z,]+/
-const chunkAttributionRegex = /[^a-z]+[:]/
+
+// endOfStatementRegex: punctuation followed by whitespace
+const endOfStatementRegex = /([^A-Za-z0-9,\s]+)\s+/g
+// attributionRegex: capitalized word -> anything -> ending with :
+const attributionRegex = /([^a-z:\s]{2,}(\s[^.:]*)?:)/g
+
+const breakpointRegex = RegExp(
```
reefdog: I believe this needs to create a new instance via `new RegExp`?

slifty: Good catch (weird that nothing sqwaked!)

PR #235, opened 2019-09-11
Meyer attribution fix

slifty:
This addresses an issue where attributions were not being parsed properly in some very rare cases.  This fix requires that attribution blocks always start with a capital letter.

This also corrects an instance of this bug from an earlier test suite which was discovered by making the fix.

Issue #167


src/server/utils/cnn.js
```diff
@@ -27,7 +27,7 @@ const attributionNameRegex = /[^a-z,]+/
 // endOfStatementRegex: punctuation followed by whitespace
 const endOfStatementRegex = /([^A-Za-z0-9,\s]+)\s+/g
 // attributionRegex: capitalized word -> anything -> ending with :
```
reefdog: This explanation should be clarified. I think it means:
`one capitalized letter -> at least one character that's not a lower-case letter or colon or whitespace -> an optional group of (unsure… whitespace followed by zero or more non-colon characters?)`

slifty: I thought that too but technically a capitalized word would have to start with a capital letter so I think actually the regex now more accurately reflects the description.

The intent wasn't to have the description regurgitate the literal regex (you can just read the regex for that) but rather the intent of the regex.

slifty: Basically it is saying "Capitalized word":  start with a capital letter and keep going (so long as you don't see anything lowercase) until you hit either a space or a `:` -> once you hit a space you have satisfied the requirement of starting with a capitalized word, so just take everything after that until you hit a colon which marks the end.

reefdog: Okay, this makes sense and I agree with documenting intent and letting the RegEx speak for itself… even if it speaks in a completely incomprehensible dialect.

PR #238, opened 2019-09-11
Add another test suite for CNN

slifty:
There was a bug identified when a sentence would end with an acronym (e.g. "U.S.")  and the portion of the acronym after the first . was attributed to the next chunk.

This bug was fixed when addressing another issue, but this commit includes the test suite for the sake of regression.

Issue #167


slifty:
This is just the inclusion of a new test suite so I'll merge without review.

PR #241, opened 2019-09-11
Break speaker name on parenthetical

slifty:
We were previously splitting speaker name from affiliation using commas, but CNN regularly uses parenthetical statements before a comma (or even including a comma within the paren) after the speaker name.

For example: `PETE BUTTIGIEG (D-SOUTH BEND, IN) PRESIDENTIAL CANDIDATE`

or `JOAN THE TRUCK DRIVER (I)`

This will properly put those parens in affiliation.  It also removes the now un-necessary section of name cleanup and updates some regression tests to reflect the real expected outputs.


src/server/utils/__test__/cnn.test.js
```diff
@@ -423,12 +423,6 @@ describe('utils/cnn', () => {
       expect(cleanSpeakerName('SONIA SOTOMAYOR'))
         .toBe('SONIA SOTOMAYOR')
     })
-    it('Should remove parentheticals', () => {
-      expect(cleanSpeakerName('JIMMY DUST (D-MA)'))
-        .toBe('JIMMY DUST')
-      expect(cleanSpeakerName('JIMMY DUST (THE MOST NOTORIOUS BOTANIST IN THE WEST)'))
```
reefdog: Bye bye, excellent parenthetical.

PR #254, opened 2019-09-15
Match up Claims table and model definitions

### LINKED ISSUE #253: Increase length of Claims fields
After nominally resolving #242, I noticed more `SequelizeDatabaseError: value too long for type character varying(255)` errors in the logs, and so went hunting. Turns out I missed some very big 255-length fields: 

```
                                                               Table "public.claims"
       Column        |           Type           | Collation | Nullable |              Default               | Storage  | Stats target | Description 
---------------------+--------------------------+-----------+----------+------------------------------------+----------+--------------+-------------
 id                  | integer                  |           | not null | nextval('claims_id_seq'::regclass) | plain    |              | 
 content             | text                     |           |          |                                    | extended |              | 
 claimed_at          | timestamp with time zone |           |          |                                    | plain    |              | 
 canonical_url       | character varying(1024)  |           |          |                                    | extended |              | 
 claim_buster_score  | double precision         |           |          |                                    | plain    |              | 
 created_at          | timestamp with time zone |           | not null |                                    | plain    |              | 
 updated_at          | timestamp with time zone |           | not null |                                    | plain    |              | 
 scraper_name        | character varying(255)   |           |          |                                    | extended |              | 
 source              | character varying(255)   |           |          |                                    | extended |              | 
 speaker_name        | character varying(255)   |           |          |                                    | extended |              | 
 speaker_affiliation | character varying(255)   |           |          |                                    | extended |              | 
```

Looks like despite [modeling the Claims fields as 1024](https://github.com/TechAndCheck/tech-and-check-alerts/blob/01664696877d3b353fceb359ae63bc7c8c9e16a1/src/server/models/Claim.js#L6-L9), our actual database migrations [just set them to `Sequelize.STRING`](https://github.com/TechAndCheck/tech-and-check-alerts/blob/01664696877d3b353fceb359ae63bc7c8c9e16a1/src/server/migrations/20190716204914-adjust-claims.js#L8), and that's where it counts.

`scraper_name` should actually remain 255, and the model should be changed to match. The other three will be adjusted to 1024 to match the model.

reefdog:
This commit adds a migration to change the length of three `claims` table columns and modifies one `Claim` model attribute definitions to match the existing column (which was, incidentally, more correct).

Previously, we had a mismatch between our Sequelize model definition for Claims and the actual table structure. (That this is possible belies a bit of a problem in Sequelize itself, but I digress.) Four of the fields were defined as 1024 characters long in the model, but used the default 255-character length in their original migrations.

One of those fields, `scraper_name`, didn’t actually need to be 1024 anyway, so this changes the model definition. The other three could arguably fit into less than 1024 in many cases (especially `source`), but there’s no harm in making them long enough for edge cases.

**You will need to run migrations after merging this change.**

Resolves #253

### DIFF
--- a/src/server/migrations/20190915180014-increase-claim-column-length.js
+++ b/src/server/migrations/20190915180014-increase-claim-column-length.js
@@ -1,39 +1,46 @@
 import { runPromiseSequence } from '../utils'
 
 module.exports = {
-  up: (queryInterface, Sequelize) => runPromiseSequence([
-    () => queryInterface.changeColumn(
-      'claims',
-      'source',
-      Sequelize.STRING(1024),
-    ),
-    () => queryInterface.changeColumn(
-      'claims',
-      'speaker_name',
-      Sequelize.STRING(1024),
-    ),
-    () => queryInterface.changeColumn(
-      'claims',
-      'speaker_affiliation',
-      Sequelize.STRING(1024),
-    ),
-  ]),
-
-  down: (queryInterface, Sequelize) => runPromiseSequence([
-    () => queryInterface.changeColumn(
-      'claims',
-      'source',
-      Sequelize.STRING,
-    ),
-    () => queryInterface.changeColumn(
-      'claims',
-      'speaker_name',
-      Sequelize.STRING,
-    ),
-    () => queryInterface.changeColumn(
-      'claims',
-      'speaker_affiliation',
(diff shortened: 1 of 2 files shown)


reefdog:
Changed, tested locally (up and down), merged, deployed, and migrations run on server.

PR #257, opened 2019-09-17
Save Twitter claims using preferred display name

### LINKED ISSUE #249: Use preferred Twitter display names
@markstencel noticed that sometimes we are possibly using Twitter handles instead of the desired display names. Filing this as a placeholder for checking this issue. @markstencel Please update with examples if you have them because I haven't been tracking this!

reefdog:
By design, we intended to associated Twitter claims with the human-curated “preferred display name” stored in `TwitterAccount.preferredDisplayName`. Previously, though, we were just using the Twitter account’s scraped name.

Now, we actually use the preferred display name from the database, if it exists and is non-empty.

Resolves #249

Note: because we introduced new references to `speaker.name`, either this branch or #268 will need adjusting based on who gets merged first.

### DIFF
--- a/src/server/utils/twitter.js
+++ b/src/server/utils/twitter.js
@@ -1,5 +1,10 @@
 import querystring from 'querystring'
 
+import { hasVisibleContent } from '.'
+import models from '../models'
+
+const { TwitterAccount } = models
+
 export const isTwitterScreenName = screenName => /^@?(\w){1,15}$/.test(screenName)
 
 export const getTwitterApiUrlForUserTimeline = (screenName) => {
@@ -35,3 +40,120 @@ export const extractStatementsFromTweets = tweets => tweets.map(tweet => ({
   canonicalUrl: getCanonicalUrlFromTweet(tweet),
   claimedAt: getTimestampFromTweet(tweet),
 }))
+
+/**
+ * Converts a screen name to a form that can reliably be used as a dict key.
+ *
+ * @param  {String} screenName The screen name you want to hash
+ * @return {String}            The screen name hash
+ */
+export const getScreenNameHash = screenName => screenName.toLowerCase()
+
+/**
+ * Returns the better of two name options.
+ *
+ * For Twitter, the `a` name will be the name scraped from the API, while the `b` name will be the
+ * name we've stored as "preferred" for that account. We don't do any fancy heuristics here,
+ * trusting in human intent. All we really do is make sure the preferred name isn't empty.
+ *
+ * @param  {String} a The default name
+ * @param  {String} b The candidate name for improvement
+ * @return {String}    The best name
+ */
+export const getBestName = (a, b) => (
+  hasVisibleContent(b) ? b : a
+)
+
+/**
+ * Given an array of statements, returns an array of screen names.
+ *
+ * @param  {Array<Statement>} statements An array of statements from which to pluck the screen name
+ * @return {Array<String>}               An array of screen names
+ */
+export const getScreenNamesFromStatements = statements => statements
+  .map(statement => statement.source)
+
+/**
+ * An accumulator function used by `reduce()` to extract an account's preferred display name and
+ * add it to the accumulator.
+ *
+ * @param  {Object}         displayNamesByScreenName A dict of display names keyed by screen name
+ * @param  {TwitterAccount} account                  The Twitter account we're adding to the dict
+ * @return {Object}                                  The dict, with the account added
+ */
+const applyDisplayNameToDict = (displayNamesByScreenName, account) => {
+  const {
+    screenName,
+    preferredDisplayName,
+  } = account
+  return Object.assign({}, displayNamesByScreenName, {
+    [getScreenNameHash(screenName)]: preferredDisplayName,
+  })
+}
+
+/**
+ * Given an array of screen names, looks up the preferred display names from the database and
+ * returns a dict mapping screen names to display name.
+ *
+ * Note that (1) this requires calling the database so it is an async functionm and (2) the dict
+ * that this returns uses screen name hashes as keys, so be sure to hash screen names through
+ * `getScreenNameHash()` when accessing the dict values that this function returns.
+ *
+ * @param  {Array<String>} screenNames An array of screen names
+ * @return {Promise<Object>}           When the Promise resolves, returns a dict mapping screen
+ *                                     names to display names
+ */
+export const getDisplayNamesForScreenNames = async screenNames => TwitterAccount
+  .getByScreenNames(screenNames)
+  .then(accounts => accounts.reduce(applyDisplayNameToDict, {}))
+
+/**
+ * Modifies the given statement with a normalized speaker name, using a collection of preferred
+ * display names (keyed by screen name).
+ *
+ * This function requires the entire statement as input since it relies on `statement.source` for
+ * display name lookup.
+ *
+ * Design caveat: Arguably, it could receive a statement but only return a speaker, but that
+ * version felt unintuitive and less legible than a statement in / statement out design.
+ *
+ * @param  {Statement} statement                The statement to apply a display name to
+ * @param  {Object}    displayNamesByScreenName A dict of display names keyed by screen name
+ * @return {Object}                             The statement, with speaker name normalized
+ */
+export const normalizeStatementSpeaker = (statement, displayNamesByScreenName) => {
+  const {
+    source,
+    speaker,
+    speaker: {
+      name: speakerName,
+    },
+  } = statement
+  const preferredDisplayName = displayNamesByScreenName[getScreenNameHash(source)]
+  const normalizedSpeakerName = getBestName(speakerName, preferredDisplayName)
+  return Object.assign({}, statement, {
+    speaker: {
+      ...speaker,
+      name: normalizedSpeakerName,
+    },
+  })
+}
+
+/**
+ * Given a list of statements, normalizes the speaker names for all the statements.
+ *
+ * In practice, this means looking up our preferred display names for each associated Twitter
+ * account from the database and applying them to each statement speaker.
+ *
+ * We want to await the database lookup, so this function is marked async.
+ *
+ * @param  {Array<Statement>} statements An array of statements whose speakers should be normalized
+ * @return {Promise<Array>}              A Promise that, when resolved, returns the array of
+ *                                       statements with normalized speaker names
+ */
+export const normalizeStatementSpeakers = async (statements) => {
+  const screenNames = getScreenNamesFromStatements(statements)
+  const displayNamesByScreenName = await getDisplayNamesForScreenNames(screenNames)
+  return statements
+    .map(statement => normalizeStatementSpeaker(statement, displayNamesByScreenName))
+}
--- a/src/server/workers/scrapers/TwitterAccountStatementScraper.js
+++ b/src/server/workers/scrapers/TwitterAccountStatementScraper.js
@@ -1,12 +1,15 @@
 import OAuth from 'oauth'
+
 import config from '../../config'
 import { STATEMENT_SCRAPER_NAMES } from './constants'
 
+import { runAsyncSequence } from '../../utils'
 import {
   isTwitterScreenName,
   getTwitterApiUrlForUserTimeline,
   parseJsonIntoTweets,
   extractStatementsFromTweets,
+  normalizeStatementSpeakers,
 } from '../../utils/twitter'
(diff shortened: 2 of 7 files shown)


src/server/workers/scrapers/TwitterAccountStatementScraper.js
```diff
@@ -44,20 +54,28 @@ class TwitterAccountStatementScraper extends AbstractStatementScraper {
     )
   }
 
-  extractStatementsFromTwitterApiResponse = (responseString) => {
-    const stepSequence = [
-      parseJsonIntoTweets,
-      extractStatementsFromTweets,
-    ] // Note that order does matter here
+  applyPreferredDisplayNameToStatements = async (statements) => {
```
slifty: My gut is that we should define this as utils (rather than class methods); though at that point we may not need a helper method for getTwitterAccount.

src/server/workers/scrapers/TwitterAccountStatementScraper.js
```diff
@@ -44,20 +54,28 @@ class TwitterAccountStatementScraper extends AbstractStatementScraper {
...
+    const twitterAccount = await this.getTwitterAccount()
+    const { preferredDisplayName } = twitterAccount
 
-    const statements = stepSequence.reduce((string, fn) => fn(string), responseString)
-    return statements
-  }
+    // We only want to transform statements using the `preferredDisplayName` if it's better than
+    // the scraped name. Without a real heuristic, we'll settle for "it exists and is non-empty".
+    const isValidDisplayName = displayName => displayName.trim().length > 0
```
slifty: Using CNN name normalization as a baseline, this kind of heuristic check has its own (tested, javadoc'd) utility method called [getBestName](https://github.com/TechAndCheck/tech-and-check-alerts/blob/01664696877d3b353fceb359ae63bc7c8c9e16a1/src/server/utils/cnn.js#L343-L345) which takes in the two options and picks the best.  In this case the two names are weighted differently (e.g. the parameters could be "defaultName and preferredName" to make it clear that the preferredName will be used if it exists)

src/server/workers/scrapers/TwitterAccountStatementScraper.js
```diff
@@ -44,20 +54,28 @@ class TwitterAccountStatementScraper extends AbstractStatementScraper {
...
 
-  statementScrapeHandler = (responseString) => {
-    const statements = this.extractStatementsFromTwitterApiResponse(responseString)
-    return statements
+    return isValidDisplayName(preferredDisplayName) ? statements.map(statement => ({
```
slifty: Take a look at the [CNN normalization](https://github.com/TechAndCheck/tech-and-check-alerts/blob/01664696877d3b353fceb359ae63bc7c8c9e16a1/src/server/utils/cnn.js#L368-L381) method which feels a lot easier to read somehow.

What do you think of having the name normalization logic go something like this:

1. Get a complete handle => display name dict (`displayNames`)
2. Map that statements in a way that runs for everything and invokes a twitter equivalent of `getNormalizedSpeaker` (e.g. `getNormalizedName(statement, displayNames)`) which basically just  invokes the earlier suggested `getBestName(statement.speaker.name, displayNames[statement.handleOrWhatever])`

The CNN version manages to avoid a ton of ...'s by using `Object.assign` which feels a lot cleaner to me.

reefdog: This makes a ton of sense. I'm going to refactor around this model. Thank you!

src/server/workers/scrapers/TwitterAccountStatementScraper.js
reefdog: I think refactoring around the CNN normalization pattern will result in this going away entirely, as mentioned.

src/server/workers/scrapers/TwitterAccountStatementScraper.js
reefdog: A question about "Get a complete handle => display name dict": that's only possible if we've got a place to store that dict that's accessible to each instance of `TwitterAccountStatementScraper`, or if we run the normalizer at a higher level (such as job processor level). Am I wrong there?

slifty: I was imagining something like this:

```
export const normalizeStatementSpeakers = (statements) => {
  const screenNames = getScreenNamesFromStatements(statements)
  const displayNames = getDisplayNamesForScreenNames(screenNames)
  return statements.map(statement => Object.assign({}, statement, {
    speaker: applyDisplayNameToSpeaker(statement.speaker, displayNames),
  }))
}
```

reefdog:
Okay, substantial change just committed per review feedback by @slifty. Re-review requested!

There were a few challenges to overcome here, not least of which Twitter's speaker name normalization, unlike CNN's, requires a trip to the database to lookup preferred display names. This means inserting an async/await into the sequence, which I'm not fond of but seems to work.

Also, this introduces some new references to `speaker.name`, which means there will be some additional sync-up once #268 is merged (or, conversely, we'll want to marge and adjust here if #268 is completed before this one).

src/server/utils/twitter.js
```diff
@@ -35,3 +39,98 @@ export const extractStatementsFromTweets = tweets => tweets.map(tweet => ({
...
+ * use this function's results.
+ *
+ * @param  {Array<String>} screenNames An array of screen names
+ * @return {Promise<Object>}           When the Promise resolves, returns an object mapping of
+ *                                     screen names to display names
+ */
+export const getDisplayNamesForScreenNames = async screenNames => TwitterAccount
+  .getByScreenNames(screenNames)
+  .then(accounts => accounts.reduce((mapping, account) => {
+    // eslint-disable-next-line no-param-reassign
```
slifty: 1) Can we use Object.assign instead of disabling a linting rule?
2) Should we put the "screen name to dict label" conversion in a method so that conversion isn't hidden inside of this method / it can be used elsewhere.  (e.g. `getScreenNameHash()`)

```
const screenHash = getScreenNameHash(account.screenName)
Object.assign({
  `${screenHash}`: account.preferredDisplayName,
}, mapping)
```

^ Note that I forget the syntax for Object.assign so treat that as pseudocode.

src/server/utils/twitter.js
```diff
@@ -35,3 +39,98 @@ export const extractStatementsFromTweets = tweets => tweets.map(tweet => ({
...
+ */
+export const getScreenNamesFromStatements = statements => statements
+  .map(statement => statement.source)
+
+/**
+ * Given an array of screen names, looks up the preferred display names from the database and
+ * returns a mapping of screen name to display name.
+ *
+ * Note that (1) this requires calling the database, so this is an async function that should
+ * generally be awaited, and (2) we convert the screen names to lowercase when composing the
```
slifty: If you move that "hashing" logic to its own thing you can remove that (2) note

src/server/utils/twitter.js
```diff
@@ -35,3 +39,98 @@ export const extractStatementsFromTweets = tweets => tweets.map(tweet => ({
...
   claimedAt: getTimestampFromTweet(tweet),
 }))
+
+/**
+ * Validates that the string has non-whitespace content.
+ *
+ * @param  {String} text The text you are curious about, o curious one
+ * @return {Boolean}     True if it has non-whitespace content, else false
+ */
+export const hasVisibleContent = text => !!text && text.replace(/\s/g, '').length > 0
```
slifty: This isn't Twitter specific, right?  If not, lets put it in a more generic namespace (either `strings.js` or `index.js` or whatever feels right).
^ NOTE: I appreciate that we aren't super intentional with our util namespaces so, feel free to say "nah" and punt that kind of thing to a never-gonna-actually-happen-refactor.

slifty: (Right now we have the squishing util in `index.js` so it probably belongs there; though we should make an issue to move both of them to `strings.js`)

src/server/workers/scrapers/TwitterAccountStatementScraper.js
```diff
@@ -44,20 +47,11 @@ class TwitterAccountStatementScraper extends AbstractStatementScraper {
...
-  }
-
-  statementScrapeHandler = (responseString) => {
-    const statements = this.extractStatementsFromTwitterApiResponse(responseString)
-    return statements
-  }
+  statementScrapeHandler = responseString => runSequence([
+    parseJsonIntoTweets,
+    extractStatementsFromTweets,
+    normalizeStatementSpeakers,
```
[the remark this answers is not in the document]
reefdog: Wow! I'm glad you knew the solution to this. It smelled, but when it worked I decided I just didn't understand computers. Thank you!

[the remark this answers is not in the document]
reefdog: (And yes, I'm guessing now that if we'd stacked a non-async function after `normalizeStatementSpeakers()` in the sequence, the seams would've ripped. In fact I'm gonna test that for kicks.)

src/server/utils/twitter.js
reefdog: `Object.assign` syntax aside, this is a good idea. I'm also going to break out the accumulator function.

BUT, unfortunately that method of dynamic key name assignment doesn't work — it was the first thing I tried too — so it'll be a little more verbose.

src/server/utils/twitter.js
reefdog: I think the note is still useful, because I want consumers of `getDisplayNamesForScreenNames()` to be abundantly clear they'll need to understand the need to hash keys when _accessing_ this dict, too.

reefdog: (I do agree I should remove the detail about what a hashed key looks like.)

reefdog:
Ready for re-review!

src/server/utils/twitter.js
```diff
@@ -35,3 +40,120 @@ export const extractStatementsFromTweets = tweets => tweets.map(tweet => ({
...
   claimedAt: getTimestampFromTweet(tweet),
 }))
+
+/**
+ * Converts a screen name to a form that can reliably be used as a dict key.
+ *
+ * @param  {String} screenName The screen name you want to hash
+ * @return {String}            The hashed screen name
+ */
+export const getHashedScreenName = screenName => screenName.toLowerCase()
```
slifty: "HashedScreenName" vs "ScreenNameHash"
^ the return value of this method happens to look like a screen name, but the reality is that it is intended to be a hash, not a screen name (you could return an md5 and it would work just as well).

In other words, the noun returned is a hash (of a screen name), e.g. a ScreenNameHash

src/server/utils/twitter.js
```diff
@@ -35,3 +40,120 @@ export const extractStatementsFromTweets = tweets => tweets.map(tweet => ({
...
+ * @param  {Object}         displayNamesByScreenName A dict of display names keyed by screen name
+ * @param  {TwitterAccount} account                  The Twitter account we're adding to the dict
+ * @return {Object}                                  The dict, with the account added
+ */
+const applyDisplayNameToDict = (displayNamesByScreenName, account) => {
+  const {
+    screenName,
+    preferredDisplayName,
+  } = account
+  const displayNameByScreenName = {}
```
slifty: I found out how to do [dynamic property names in the {} structure itself](https://stackoverflow.com/a/35579786) (which will be more succinct)

```
Object.assign(
  {},
  displayNamesByScreenName,
  {
    [getScreenNameHash(screenName)]: preferredDisplayName,
  },
)
```

src/server/workers/scrapers/TwitterAccountStatementScraper.js
```diff
@@ -44,20 +47,11 @@ class TwitterAccountStatementScraper extends AbstractStatementScraper {
...
-
-    const statements = stepSequence.reduce((string, fn) => fn(string), responseString)
-    return statements
-  }
-
-  statementScrapeHandler = (responseString) => {
-    const statements = this.extractStatementsFromTwitterApiResponse(responseString)
-    return statements
-  }
+  statementScrapeHandler = responseString => runAsyncSequence([
```
slifty: This should now be explicitly marked as async

`statementScrapeHandler = async responseString => ...`

src/server/utils/twitter.js
reefdog: Bless you.

reefdog:
All done! Wanna glance?

reefdog:
Neat! I’ll squash and merge later this afternoon.

slifty:
wait!

Issue #100, opened 2019-08-01
Scrape "Claimed At" time for CNN


slifty:
We want to try to get the time a claim was made (resolution of program start time is fine for now, but as specific as possible given the transcript is what we should ultimately aim for).

This will be a matter of finding the date + time on the transcript page; worst case we can use the date in the URL but that will only give us resolution of the day, not the hour.

reefdog:
Quick thought that we'll want to make sure we sniff out the timezone for the timestamp, too, and if we're doing anything like the worst case (URL date) that we're localizing it to UTC and considering how that will affect our queries.

slifty:
So!  Twitter this should be really easy since I think twitter does UTC.

CNN, however, seems to write "ET" which isn't even an actual time zone.  I'm honestly a little flummoxed as to what to do, especially since historic scrapes make the problem 2x harder (since EST / EDT will change based on the day) AND we don't actually care about historic scrape.

Options are:

1) Convert to UTC based on what the time would be if it was aired at the "ET" time *today*
2) Try to be smarter to figure out whether it is EST or EDT based on the air date.

#2 is the more correct option, but I'm not sure it is worth the effort it will take.

reefdog:
I think you mean (B) is the more correct but harder option? (And assuming we're converting to / storing UTC in both options, anyway.)

I can't believe everyone isn't using a `<time>` element with a valid ISO 8601 `datetime` attribute. Tsk tsk.

slifty:
I suppose I do mean B if you want to be exact about it.

reefdog:
If at all possible.

slifty:
OK so I spent some time yesterday digging into this -- yikes what a nightmare.

Some things worth putting down as a baseline (which folks probably already know):

1. Javascript's native Date object does not do time zones
2. `dayjs` unfortunately does not do time zones
3. `moment` does, but it's so huge (RIP)
4. [Luxon](https://moment.github.io/luxon/index.html) is a spinoff of moment which is much smaller and also has timezone support.
5. There's this [spacetime](https://github.com/spencermountain/spacetime) thing which is explicitly for moving between time zones.

I'm not sure we can get around the need for another third party at this point.  Eventually it looks like [dayjs will have time zone support](https://github.com/iamkun/dayjs/issues/46) but it isn't there yet.

I'm not sure the value of dayjs if we have to use another library to supplement it.

reefdog:
[...]

PR #259, opened 2019-09-17
Support `Claim.claimedAt` and use it for Twitter

### LINKED ISSUE #258: Scrape "Claimed At" time for Twitter
We've always planned to store the time a claim was made (or more accurately, published) as `Claim.claimedAt` and use this, rather than [the time we stored the claim](https://github.com/TechAndCheck/tech-and-check-alerts/blob/01664696877d3b353fceb359ae63bc7c8c9e16a1/src/server/newsletters/NationalNewsletter.js#L56), for our newsletter-generation queries.

For Twitter, this is so easy (the API serves it to us on a silver platter) that we needn't wait until CNN's trickery is resolved (#100).

reefdog:
We’ve had a `Claim.claimedAt` attribute, but haven’t been using it. Instead, we've been using `createdAt` as a proxy.

This PR adds the infrastructure for scrapers to populate `Claim.claimedAt`; modifies newsletters to query by `claimedAt` rather than `createdAt`; and updates the Twitter scraper to populate the field. It will take longer to reliably scrape `claimedAt` for CNN (see #100), so those will continue to fall back to the scrape time instead.

After deploying, we should run `UPDATE claims SET claimed_at = created_at WHERE claimed_at IS NULL` for backwards-compatibility.

Affects #100
Resolves #258

### DIFF
--- a/src/server/newsletters/NationalNewsletter.js
+++ b/src/server/newsletters/NationalNewsletter.js
@@ -53,7 +53,7 @@ class NationalNewsletter extends AbstractNewsletter {
   generateQueryParams = where => ({
     where: {
       ...where,
-      createdAt: {
+      claimedAt: {
         [Sequelize.Op.gte]: dayjs().startOf('hour').subtract(1, 'day').format(),
         [Sequelize.Op.lt]: dayjs().startOf('hour').format(),
       },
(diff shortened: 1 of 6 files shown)


reefdog:
Merged, deployed, and ran `UPDATE claims SET claimed_at = created_at WHERE claimed_at IS NULL` to ensure existing `claimed_at` columns fell back to `created_at`.

PR #260, opened 2019-09-23
Update README

emilyliu7321:
This commit provides additional instructions on setting up PostgreSQL and Redis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ This is a node project, to set up your development environment...
 
 ### Set up postgres
 
-This project uses [postgres](https://www.postgresql.org/) to store its data.  You will need to either set up a local copy, or find a hosted solution.
+This project uses [postgres](https://www.postgresql.org/) to store its data.  You will need to either set up a local copy, or find a hosted solution. Be sure to remember the username and password you set up for this project. 
 
 ### Set up redis
 
@@ -29,6 +29,8 @@ You will need to configure your environment variables.  In production this can b
 > cp .env.template .env
 > vi .env
 ```
+
+The `DATABASE_URL_DEVELOPMENT` template should match the `username` and `databasename` that you set up when installing postgres. 
 
 ### Run Migrations
 


README.md
```diff
@@ -15,12 +15,60 @@ This is a node project, to set up your development environment...
 
 ### Set up postgres
 
-This project uses [postgres](https://www.postgresql.org/) to store its data.  You will need to either set up a local copy, or find a hosted solution.
+This project uses [postgres](https://www.postgresql.org/) to store its data.  You will need to either set up a local copy, or find a hosted solution. 
+
+To install PostgreSQL, run
+```
+brew install postgresql
```
slifty: This is an OSX specific command, we may want to keep or README platform agnostic.

README.md
```diff
@@ -15,12 +15,60 @@ This is a node project, to set up your development environment...
...
+postgres=#
+```
+Now, run this, where you should change `username` to the user that you want to create.
+```
+CREATE USER username;
+```
+
+This user should be able to create a DB, so set the role to superuser. 
+```
+postgres=# ALTER USER username SUPERUSER CREATEDB;
```
slifty: Do we need the user to be a superuser for this project outside of creating the database?

I would suggest that the database be created by the root user (not the `alerts` user) and that the root user also grant access to the database.  This is safer since the alerts user (which will be credential'd by code) would not have superuser access (thus if the application is compromised by a bug or a bad actor the vulnerability would be limited to alerts data, not the entire psql installation)

README.md
```diff
@@ -30,6 +78,8 @@ You will need to configure your environment variables.  In production this can b
 > vi .env
 ```
 
+The `DATABASE_URL_DEVELOPMENT` template should match the `username` and `databasename` that you set earlier.
```
slifty: I think this would be a reasonable note to include here regardless of changes to psql documentation (maybe slightly modified to something like `... that you set up when installing postgres` depending on how we change the earlier documentation)

reefdog:
Agree with @slifty on all accounts, which should make me question my life choices.

This is well-written and I have a particular affection for the care in the "little" things like accurate prompts and good Markdown formatting. Philosophically, though, I agree that some of this detail could go in a platform-specific setup instructions (if documenting for other devs) or the [wiki](https://github.com/TechAndCheck/tech-and-check-alerts/wiki) (if documenting for history / working out how to script).

There's a fine line, because commands like `yarn install` can/should be documented (they're part of the project setup, not necessarily environment setup). We'll feel our way toward the right balance!

Either way, :tada: for the first PR and thank you for your care in it!

README.md
```diff
@@ -15,12 +15,60 @@ This is a node project, to set up your development environment...
...
 ### Set up postgres
 
-This project uses [postgres](https://www.postgresql.org/) to store its data.  You will need to either set up a local copy, or find a hosted solution.
+This project uses [postgres](https://www.postgresql.org/) to store its data.  You will need to either set up a local copy, or find a hosted solution. 
+
+To install PostgreSQL, run
+```
+brew install postgresql
+```
+Create a local user and database. More instructions can be found [here](https://www.postgresql.org/docs/8.0/sql-createuser.html).
```
reefdog: It's a small thing, but consistent line breaks throughout the docs would be good. I think a newline after each code block, which you start doing later on, would be good.

PR #262, opened 2019-09-24
Remove jobs on completion by default

### LINKED ISSUE #261: Remove jobs on completion
It looks like our jobs never get removed from Redis, which means that over time the redis store gets bigger and bigger and bigger an...

Bull has a [remove on completion](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queueadd) parameter when a job is added.  Lets update our AbstractJobScheduler to pass in `removeOnComplete: true`

Note: we could choose to pass an integer value instead of true and it bull would keep a smaller history of completed jobs.  I'm not sure we get much value from this, but is there any value in supporting that in a dev environment?

slifty:
When the scrapers were running we were slowly accumulating jobs in redis which over time essentially served as a memory leak.  This updates our queues to default to deleting jobs upon completion.

Resolves #261

### DIFF
--- a/src/server/queues/AbstractQueueFactory.js
+++ b/src/server/queues/AbstractQueueFactory.js
@@ -37,6 +37,7 @@ class AbstractQueueFactory {
       {
         defaultJobOptions: {
           removeOnComplete: true,
+          removeOnFail: true,
         },
       },
     )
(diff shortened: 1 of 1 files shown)


slifty:
Good call @reefdog  -- I'll just set to true for now; removeOnFail does strike me as maybe having a bene

slifty:
Whoops I didn't mean to close...

PR #265, opened 2019-10-02
Add more robust TESTING.md documentation

### LINKED ISSUE #47: Populate TESTING.md

slifty:
This attempts to capture the practices we have invoked so far.

There are still some areas that are not fleshed out (e.g. integration testing instructions) but there is no need to invent that until we actually begin to write integration tests.

Resolves #47

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -6,15 +6,68 @@ We aspire to have a full-blown robust unit and integration test suite, but for n
 
 Still: if you write code that can be tested, please ensure all PRs include tests of that code.
 
-## Generating tests for the CNN scraper
+## Test organization
+
+### Unit Tests
+Unit tests should be placed in a __test__/ directory that lives at the same level of the file being tested.  Each file should have a parallel `.test.js`
+
+For example: 
+
+```
+|- /src
+|  |- __test__
+|  |  |- sillyHat.test.js
+|  |- sillyHat.js
+```
+
+### Integration Tests
+
+Since integration tests often involve multiple files, they should live in a common `src/tests` directory.  We have not yet written any integration tests so this documentation should be updated when that occurs.
+
+### Test Data
+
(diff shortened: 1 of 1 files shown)


docs/TESTING.md
```diff
@@ -6,15 +6,68 @@ We aspire to have a full-blown robust unit and integration test suite, but for n
 
 Still: if you write code that can be tested, please ensure all PRs include tests of that code.
 
-## Generating tests for the CNN scraper
+## Test organization
+
+### Unit Tests
+Unit tests should be placed in a __test__/ directory that lives at the same level of the file being tested.  Each file should have a parallel `.test.js`
```
reefdog: `__test__/` renders as "**test**/", unfortunately. Need to escape these.

slifty: This would have been good to know 4 minutes ago

reefdog: Yeah, you couldn't have waited four more minutes?

slifty: Obviously YOU could.

docs/TESTING.md
```diff
@@ -6,15 +6,68 @@ We aspire to have a full-blown robust unit and integration test suite, but for n
...
+|  |- sillyHat.js
+```
+
+### Integration Tests
+
+Since integration tests often involve multiple files, they should live in a common `src/tests` directory.  We have not yet written any integration tests so this documentation should be updated when that occurs.
+
+### Test Data
+
+Test data should be placed in a file that is sibling to the test file and have the `TestData.js` prefix.
```
reefdog: `prefix` → `suffix`

PR #268, opened 2019-10-03
Avoid reserved words

### LINKED ISSUE #223: Avoid reserved words in Speaker model properties
There's a [list of very reserved and mostly reserved words in JavaScript](https://www.w3schools.com/js/js_reserved.asp) that we should avoid using for the names of variables and object properties, if at all possible. Our [Speaker model](https://github.com/TechAndCheck/tech-and-check-alerts/blob/d0dba888a7fdc497b9ac530df51b2fad49494360/src/server/utils/cnn.js#L9-L11) has a `name` property; we should change this to `speakerName` or `fullName` or something.

emilyliu7321:
This renames the `name` property in our Speaker model to `fullName` to avoid using a Javascript reserved word. This also modifies the corresponding test file to check for `fullName` instead of `name`.

Resolves #223


emilyliu7321:
1. Yes, I looked for instances of [Javascript reserved words](https://www.w3schools.com/js/js_reserved.asp) to check if any were improperly used. `name` is the only one I found!   
2. Fixed -- is there an easy way to check for mistakes like that while integration tests are being written? 
3. Modified `Claim.speakerName` to `Claim.speakerFullName` but like the case above, not sure if passing all of the tests now really covers 100% of cases

src/server/migrations/20190716204924-remove-speakers.js
```diff
@@ -98,7 +98,7 @@ module.exports = {
         UPDATE claims
         SET    speaker_id = speakers.id
         FROM   speakers
-        WHERE  speakers.full_name = claims.speaker_name
+        WHERE  speakers.full_name = claims.speaker_full_name
```
reefdog: We should undo these changes to this migration file and instead create a new migration file.

slifty: (Justin this is a reply to your overall review not this specific comment)

TL;DR

reefdog: I completely agree. I refused to even read it while writing it!

slifty: ... what are you talking about right now?

reefdog: @slifty Understand if you're confused. If you want in on the pairing too, I can give you a 101 on migrations.

slifty:
😬 so I am NOT SUGGESTING we step back from the full_name path that has been followed at this point but I will say that I am bummed that javascript has `name` as a reserved word...

Fine I'll just explore this out loud: full name feels inaccurate.  The speaker name may be a full name, but as we have seen it may also be the name of an organization, or maybe a nickname.  Should full name include middle name?  Oh no [there's an entire wikipedia article about full names](https://en.wikipedia.org/wiki/Personal_name).  DO WE REQUIRE FEUDAL NAMES IN OUR FULL NAME FIELD?!?!

Isn't it somewhat just circumstance that this field is normalized to full name?  I could imagine a world in which the requirement is for the name field to be normalized to last_name for instance.

If we stick with full_name (which obviously we will do since `I am NOT SUGGESTING we step back from the full_name path...`) then I do think it is too verbose to update every utility method from Name to FullName...

That said, what makes a name a name there and a full name somewhere else?

Just to casually drop some alternatives to full_name (which `I am NOT SUGGESTING we consider`):

- speaker_name
- moniker (lol but actually though...? lol no what even is a moniker 🐒.)
- identity
- display_name
- normalized_name (and a sister attribute: denormalized_name or extracted_name)
- Apple Inc.

In seriousness what do you all think of normalized_name and extracted_name.

reefdog:
Quick note @emilyliu7321: I just merged in #249, which introduced a few new instances of `speaker.name` that we need to handle. 😬 Sorry about that! I think this is a complete list of locations that'll need updating:

https://github.com/TechAndCheck/tech-and-check-alerts/pull/257/files#diff-9b769717b9d16e9c2adecc53c735f953R3
https://github.com/TechAndCheck/tech-and-check-alerts/pull/257/files#diff-9b769717b9d16e9c2adecc53c735f953R14
https://github.com/TechAndCheck/tech-and-check-alerts/pull/257/files#diff-d69229d239ac0c8ec040fbf42e623c12R129
https://github.com/TechAndCheck/tech-and-check-alerts/pull/257/files#diff-d69229d239ac0c8ec040fbf42e623c12R137

emilyliu7321:
Ready for re-review!

slifty:
We talked about this a bit in Slack, and I want to convey it here.  I think this will also answer some of @reefdog's questions

The earlier proposal for splitting speaker.name into speaker.extractedName and speaker.normalizedName was not intended to affect the claim model (`claim.speakerName` still feels like an appropriate field, rather than trying to have the claim model artificially mirror the way we store data in the currently-transient speaker object).

In an ideal world we would be normalizing speakers in our schema, having a claim.speakerId that can go back and reference a speaker object.

That said, in that same ideal world we might also want to implement a `getBestName()` method in the speaker model which would allow claim-based use cases to gracefully ignore the internal name types unless the use case demanded one or the other it for some reason.

The point is, I would offer that `claim.speaker.getBestName()` is the actual ideal, but that it is is out of the scope for this issue.

However, I think that having our claims explicitly store those speaker-object internal name types is coupling claims to speakers too much.  Lets only store what we want in claim -- which is speakerName.

---

TL;DR: can we NOT change the claim model, but instead just set the claim.speakerName field to be the result of this utility method: `getBestSpeakerName = speaker => speaker.normalizedName | speaker.extractedName`

emilyliu7321:
Is using the utility method syntax necessary? Instead, I am thinking of simply changing claimBusterClaimDetectorJobProcessor.js to replace `speakerExtractedName` and `speakerNormalizedName` with 
`speakerName: claim.speaker.extractedName | claim.speaker.normalizedName, ` 

(Lines 11-12)

slifty:
Necessary is a strong word, but some reasons why putting it in a method is a good idea:

1) Using a named function makes the purpose explicit without requirng documentation (what are you trying to accomplish with that phrase?  You're looking to get the best speaker name)
2) Putting that in a method allows you to test that method (make sure that the intent of returning an extractedName as the bestName only happens in the event that normalizedName doesn't exist)
3) It allows us to use that logic any time we want to get the best name from a speaker.

emilyliu7321:
created speaker.js and speaker.test.js files 
removed changes to claim model 

ready for review

PR #271, opened 2019-10-11
Move Twitter test data to Twitter test data file

### LINKED ISSUE #270: Move Twitter test data to twitterTestData file
While doing #257, I created a file for Twitter test data so it's not threaded into the test file itself. Need to move remaining test data (unrelated to that PR) in.

(This is a quick fix that I actually performed while doing #257, but then split off for purity.)

reefdog:
In #257, we added a file for Twitter test data, but only used it for data relevant to that PR. This PR completes the transition by moving remaining test data out of the test file itself.

Resolves #270

### DIFF
--- a/src/server/utils/__test__/twitterTestData.js
+++ b/src/server/utils/__test__/twitterTestData.js
@@ -1,3 +1,25 @@
+export const tweet = {
+  id_str: '12345',
+  full_text: 'Text',
+  user: {
+    name: 'Name',
+    screen_name: 'screenName',
+    description: 'Bio',
+  },
+}
+export const tweets = [tweet]
+
+export const extractedStatement = {
+  speaker: {
+    name: tweet.user.name,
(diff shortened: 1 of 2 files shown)


src/server/utils/__test__/twitterTestData.js
```diff
@@ -16,6 +38,19 @@ export const normalizedStatement = Object.assign({}, statement, {
 })
 export const normalizedStatements = [normalizedStatement]
 
+export const validScreenNames = [
+  'slifty',
+  'sLiFtY',
+  '@slifty',
+  'justin_is_lame',
```
slifty: I don't remember if I wrote this originally but either way I approve of it.

reefdog: You did and I'm angry I forgot to change it in this PR but too committed to the process to do something like force-push now.

Issue #275, opened 2019-11-15
Set up continuous deployment for staging (dev)


slifty:
As part of #272 we would like to have a continuous deployment to a staging server ("dev") that would reflect any code that has been pushed to the master branch.

slifty:
Planning to use something along these lines for now: https://florianbrinkmann.com/en/deployment-travis-ci-3485/

Since we only have single servers hosting this code, this should work just fine.  We can make things more complicated some day if that bridge needs to be crossed.

slifty:
This is a useful resource as well: https://github.com/dwyl/learn-travis/blob/master/encrypted-ssh-keys-deployment.md

slifty:
Closed by #279

PR #279, opened 2019-12-04
275 add dev cd

slifty:
This PR adds support for automatic deployment to a development server when a commit is made to a particular branch (the branch is configured in Travis, but we will have it be master according to the discussion in #275)

The deployment involves:

1. Uploading the latest files (delivery.sh)
2. Running migrations / installing dependencies
3. Restarting the pm2 server

These scripts won't really make sense to be run outside of the context of Travis and continuous deployment, though at some point we may want to modify them to allow a local user to deploy to dev manually using the same scripts as Travis.


deploy_key.pub
```diff
@@ -0,0 +1 @@
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQCjCz5Mk5D87fFL79PAiBrZSwUjmgVFXRRxX6QiSDpNH7lEVMFMHt6BWDJTAa0nJjUsUSFm2DEFghjups+Uwx64q/vc7+rN5TmE72zDWrGNMcEvtmkLNAFCnOmR3jf17GRSPGX9pZ3ZZMOl8sNfOCyeTBW8udJ8bb1fTzPashUyQ2hihgRIoiJ1QFMtixCsFBsWY1bI/5X3Hs2z5xd3suqHo2N6uWGyAWq1M7PfLQX+uupKYamb+3d1UlpHjm5G8WbSpP2Ow0bmnSr/SdY4V7Y+eO8lujufUVrf1qbFyMYlEJp1Vol62BVpi3VJwEh0GK5tgTviwosMHaRxH5kIRGaVGzG8AnVSyf3QKNuI+pxRqQSm1msAnkzkfom3cHzkn61GRoQWE9r4vK47Jkr8EHLCn+ArjAd5zh7S1fX2rn15GF/NiqiUxokFf6T6umx9OlRGO20l9an/enbZJoBhfJ1ybowqq5Ps752zusT5nUanAvjY8W4Nz+JEF5wWga/Xk9N5yOKQBoTNa0F6wPbi8EEeoBBHviD8MwbzOi3ymUt4jO8DOUR7pC1mCg9DV7JyrruhA/ubYyt4U7WQjArjmTdkTutfhUxOYUAUPEDXMj8Ej1tEgYtWkVrz4RG5lCKpoYvxcgy1ecUlaiFxyGHoA0D/hO33rYp668dYo4Bz2OFsZQ== TravisCIDeployKey
```
cguess: Should this be in the repo?

.travis.yml
```diff
@@ -1,8 +1,33 @@
 language: node_js
-node_js:
-  - 10
-  - node
+
+addons:
+  ssh_known_hosts:
+  - "$STAGING_SERVER"
```
cguess: Perhaps this should be documented on how/why/what it should be set to.

.travis.yml
```diff
@@ -1,8 +1,33 @@
...
+
+before_script:
+  - echo "Host $STAGING_SERVER" >> ~/.ssh/config
```
cguess: See above.

.travis.yml
```diff
@@ -1,8 +1,33 @@
...
+  - ssh-add ./deploy_key
+
+node_js:
+  - 10
+  - node
+
+deploy:
+  - provider: script
+    script:
+      bash scripts/deliver.sh -h $STAGING_SERVER -u $STAGING_SERVER_USER bash &&
```
cguess: Same as above.

scripts/deliver.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
cguess: Can you add a bit more description on what this does? I always consider bash to be a "write once" language, but some description would be helpful.

deploy_key.pub
slifty: Yes -- it's the public key pairing to the encrypted key; you would use it on whatever server you want to deploy to.

.travis.yml
slifty: 💯 I'll add to the README.md to explain what is actually going on with all of this!

Issue #280, opened 2019-12-12
Implement "priority" speaker list


slifty:
As a result of the experimentation of #103 we are interested in improving the newsletter generation logic to select claims made by a pre-identified list of speakers.  There is plenty of room for iteration on this concept, but this issue represents the first round.

We would like to create the following:

1. A new `known_speakers` table which will contain `first_name` and `last name` fields (eventually this table may be expanded to include things like source, priority rank, etc.  For now we can keep it simple).

2. Populates that table with a seed list of known names, using a seed file.  (this should be replaced with a new worker / queue that synchronizes with a google spreadsheet.  Eventually *that* should be replaced with a more robust canonical source such as [DBPedia](https://wiki.dbpedia.org/))

3. Modify the national newsletter query to return the top 15 claims using the following logic:

* SELECT 15 claims made by known speakers (descending claim score order)
* If there are not 15 claims, fill out the rest with claims made by other speakers.

This can be accomplished through a single SQL statement using JOINs and UNIONs, but if it seems cleaner to run two separate queries and combine the results that's probably fine too.

Issue #229, opened 2019-09-10
Sync Twitter account lists from Google Sheet


reefdog:
The old system allows the team to maintain the list of desired Twitter accounts in a Google Sheet, and [fetches it at scrape time](https://github.com/TechAndCheck/tech-and-check-alerts-2018/blob/ab1bb599319f034dbe300d2890b1638df764d89e/twitter-scraper/tweet_buster.py#L32-L39). The new system stores the desired accounts locally in Postgres, and we plan to periodically update this local list from the same Sheets. (It's like a different scrape pipeline, hey neat!)

Note that the same account could be in multiple lists (say, National and North Carolina) and so any action we take (adding, deactivating, reactivating) should be specific to their presence/absence on that specific list.

reefdog:
According to @chriszs on Slack, if we don't mind the core Google Sheet being publicly-readable, a much faster way of fetching data from it than negotiating Google OAuth would be to use the sheet's public CSV URL. There's some good chatter about it on [this Stack Overflow question](https://stackoverflow.com/questions/33713084/download-link-for-google-spreadsheets-csv-export-with-multiple-sheets).

slifty:
We have two decisions to make about the google side of this issue.

# Decision 1: Nailing down the google sheet standard

Right now we have two formats for these twitter lists (one for North Carolina and one for National).  It looks like North Carolina is a bit more robust than National, so I would recommend we use that format for both.  Here are the north carolina headers:

<img width="1168" alt="image" src="https://user-images.githubusercontent.com/208884/71691299-f64e3c00-2d74-11ea-9105-d9f35045e8e5.png">

We may want also to separate party from the candidate name, creating a new column for "party".  We also might want to separate the honorific (`Sen.` and `Rep.`) from the name and put that in its own field.

The scraper really only cares about having access to:

* A clean full name (what will appear in the newsletter)
* A list of twitter handles (can be more than one, though we have to know how many columns / which columns to look at)

# Decision 2: what lookup approach should I use

Once the schema is locked in I have an implementation choice between looking up based on either:

1. index (e.g. "Coumn 3 is the twitter handle")
2. column name (e.g. "The column named Twitter is the twitter handle").

If we use the index then we cannot shift rows around / we can only add rows to the *end* of the spreadsheet if over time.

If we use column name then we can't rename column names in future / we need to be sure multiple columns don't share the same name.

chriszs:
[...]

PR #284, opened 2020-01-02
Add google spreadsheet scraper

slifty:
This creates the ability to scrape a google spreadsheet, this worker will be used in the queues created for #229 and #280 since google spreadsheets are the initial mechanism for updating things like twitter and priority speaker lists.

This creates a new dependency on a utility package called [csv-parse](https://www.npmjs.com/package/csv-parse) (which I think is necessary because please do not make me write my own CSV parser.)

### DIFF
--- a/src/server/utils/__test__/google.test.js
+++ b/src/server/utils/__test__/google.test.js
@@ -1,0 +1,16 @@
+import {
+  getSpreadsheetCsvUrl,
+  parseCsvString,
+} from '../google'
+
+describe('getSpreadsheetCsvUrl', () => {
+  it('Should return a google URL for an ID', () => {
+    expect(getSpreadsheetCsvUrl('a-google-document-id'))
+      .toBe('https://docs.google.com/spreadsheets/d/a-google-document-id/export?format=csv')
+  })
+})
+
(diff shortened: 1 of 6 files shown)


src/server/utils/__test__/google.test.js
```diff
@@ -0,0 +1,16 @@
+import {
+  getSpreadsheetCsvUrl,
+  parseCsvString,
+} from '../google'
+
+describe('getSpreadsheetCsvUrl', () => {
+  it('Should return a google URL for an ID', () => {
+    expect(getSpreadsheetCsvUrl('1gLkx2LK3yhS-glpsktWYNqBc9H2zsd7C6TvmgWjL5Kg'))
```
cguess: Just making sure, we're good with hard coding this?

slifty: it's just a test so any string would work -- given that maybe I'll just do a random string in case that one is supposed to be secret at all.

reefdog:
It's too late for me to review this but let me just say that it's _badass_ how easily the abstract scraper was able to be turned into a fully-working Google Sheet scraper. ❤

PR #286, opened 2020-01-02
Update parcel to 1.12.4

### LINKED ISSUE #285: Update parcel
Github identified a vulnerability in safer-eval and [even though parcel doesn't use the vulnerable method, it still flags as a vulnerability for projects using parcel](https://github.com/parcel-bundler/parcel/issues/3514)

I believe the parcel maintainers fixed the issue, though I am not sure the fix has been released yet.  Once it is, we should upgrade the offending package.

slifty:
This patch primarily is being made to get rid of the security
vulnerability notification on github.  The vulnerability didn't actually
affect our code base since the relevant method in the sub-library was
not being invoked by parcel.  Nevertheless, security warnings look scary
and are worth addressing even if they are inaccurate!

This also updates the lockfile to use the latest babels that fit our package versions on the next install (though we may want to consider locking those into specific versions to prevent these sorts of indirect updates whenever yarn install is run)

Resolves #285

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "eslint-plugin-react": "^7.12.4",
     "jest": "^24.8.0",
     "nodemon": "^1.19.0",
-    "parcel": "^1.12.3",
+    "parcel": "^1.12.4",
     "sequelize-cli": "^5.5.0"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


slifty:
I'm going to just merge this since it is a basic dependency upgrade.

slifty:
image

reefdog:
that's only one

slifty:
(and it is not a gif)

reefdog:
admin plz ban

Issue #252, opened 2019-09-13
Fetch read-only data in raw mode


reefdog:
When you fetch data with Sequelize, by default the objects returned are instantiated as special Sequelize objects so you can perform CRUD operations and such. For most of our data-fetching needs, we don't need this; we just want the raw data. [Thus, the `raw` option](https://sequelize.org/master/manual/models-usage.html#raw-queries) (not to be confused with [writing raw SQL](https://sequelize.org/master/manual/raw-queries.html)).

We should set this option to true whenever possible, like when fetching claims for newsletters.

reefdog:
I just spent a few minutes playing around with this, and while I still suspect it's worth looking into, the Sequelize docs just aren't clear enough about it, and it's frustratingly difficult that they reused the word "raw" to mean "writing SQL directly" and also "not wrapping the results of a query in a DAO".

Anyway, supposedly the promise of Sequelize is that if you do something like `const person = Person.create({…})`, your `person` variable is now a DAO you can use for further CRUD. We aren't using that capability, so as far as we care, the data returned by Sequelize should be raw objects. That's [the promise of `raw`](https://sequelize.org/v5/class/lib/sequelize.js~Sequelize.html#instance-method-query), as I understand it. And presumably, not wrapping it into a DAO means better performance and a little more safety (no accidentally modifying data).

I did try smashing this into our Sequelize config by adding `query: { raw: true }` to the options block for each environment (as essentially [suggested on SO](https://stackoverflow.com/a/26228558)). Nothing broke and tests all passed, but there's just enough fog around this I'm not confident committing it.

So, just gonna ice this.

PR #288, opened 2020-01-10
Add priority speakers

### LINKED ISSUE #281: Populate priority speaker list via google spreadsheet
As we work on #280 we should also invoke the same structure that needs to be created for #229 -- in other words, regularly synchronize the priority speaker list with a google spreadsheet so that we can easily add / remove names from the list.

We would want a worker + regularly-recurring queue job to invoke the sync.

We recently [learned](https://stackoverflow.com/questions/33713084/download-link-for-google-spreadsheets-csv-export-with-multiple-sheets) that you don't actually need to invoke the google API if the spreadsheet is public, which should make this an easier task.

slifty:
This creates a new queue (run every day at midnight) which scrapes a google doc and stores the results to a database.

There is a new KnownSpeaker model and a new KNOWN_SPEAKERS_GOOGLE_DOC_ID environment variable which will need to be populated on dev and prod.

Resolves #281


src/server/queues/knownSpeakerScraperQueue/knownSpeakerScraperJobProcessor.js
```diff
@@ -0,0 +1,46 @@
...
+)
+
+const extractKnownSpeakers = rows => rows
+  .filter(isValidKnownSpeakerRow)
+  .map(row => ({
+    firstName: row.first_name,
+    lastName: row.last_name,
+  }))
+
+const clearKnownSpeakers = async transaction => KnownSpeaker
```
slifty: I wonder if this should be a model method

src/server/queues/knownSpeakerScraperQueue/knownSpeakerScraperJobProcessor.js
```diff
@@ -0,0 +1,46 @@
...
+  const documentId = config.KNOWN_SPEAKERS_GOOGLE_DOC_ID
+  logger.info(`Syncing known speakers from google doc: ${documentId}`)
+  const scraper = new GoogleSpreadsheetScraper(documentId, true)
+  const rows = await scraper.run()
+  const knownSpeakers = extractKnownSpeakers(rows)
+  logger.info(`Total speakers found: ${knownSpeakers.length}`)
+
+  return sequelize.transaction(async (t) => {
+    await clearKnownSpeakers(t)
+    return knownSpeakers.forEach(async knownSpeaker => saveKnownSpeaker(knownSpeaker, t))
```
reefdog: Sequelize has [bulk insert methods](https://sequelize.org/master/class/lib/query-interface.js~QueryInterface.html#instance-method-bulkInsert) that I think we should use here instead.

src/server/queues/knownSpeakerScraperQueue/knownSpeakerScraperJobProcessor.js
```diff
@@ -0,0 +1,46 @@
...
+  }, transaction)
+
+const saveKnownSpeaker = (knownSpeaker, transaction) => KnownSpeaker.create({
+  firstName: knownSpeaker.firstName,
+  lastName: knownSpeaker.lastName,
+}, transaction)
+
+export default async () => {
+  const documentId = config.KNOWN_SPEAKERS_GOOGLE_DOC_ID
+  logger.info(`Syncing known speakers from google doc: ${documentId}`)
```
reefdog: omg tiniest thing can we capitalize `Google Doc`?

src/server/migrations/20200108172037-create-known-speakers.js
```diff
@@ -0,0 +1,39 @@
...
+
+module.exports = {
+  up: (queryInterface, Sequelize) => queryInterface.sequelize
+    .transaction(t => runPromiseSequence([
+      () => queryInterface.createTable(
+        'known_speakers',
+        {
+          id: {
+            allowNull: false,
+            autoIncrement: true,
```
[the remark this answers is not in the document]
slifty: NOW you ask.

Yes, I imagine so.  Primary keys don't necessarily need to auto increment (e.g. if you had it as a string) so I do imagine that is needed (back in the day when creating a table with raw SQL I would have to explicitly state auto increment as well as primary key).

Allow null, however, is probably redundant.

Either way I'm looking it up because now I'm curious!

src/server/queues/knownSpeakerScraperQueue/knownSpeakerScraperJobProcessor.js
```diff
@@ -0,0 +1,46 @@
...
+import config from '../../config'
+
+const {
+  sequelize,
+  KnownSpeaker,
+} = models
+
+const isValidKnownSpeakerRow = row => (
+  Object.prototype.hasOwnProperty.call(row, 'first_name')
+  && Object.prototype.hasOwnProperty.call(row, 'last_name')
```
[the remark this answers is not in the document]
reefdog: I was moving fast and didn't explain why this is more than just syntactic sugar.

First, syntactic sugar is great! Stop hatin! But mostly:

1. Verifying presence of object keys has historically been buggy with weird edge problems (though us targeting a single engine, Node, helps tremendously). Centralizing it in a utility function gives us a central location to test and sort out bugs. (That said, I should've added tests for this and didn't. `SHAME`)
2. `Object.prototype.hasOwnProperty` is totally the Standard Right Way, but I still find it absurdly illegible.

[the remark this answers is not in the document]
slifty: Well now you've convinced me

.
.

TO REJECT YOUR PROPOSED CHANGE OUT OF SPITE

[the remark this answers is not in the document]
reefdog: I suspected this would happen.

src/server/migrations/20200108172037-create-known-speakers.js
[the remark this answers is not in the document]
slifty: Looks like allowNull is not generally invoked alongside primaryKey and so I suspect is redundant.

src/server/queues/knownSpeakerScraperQueue/knownSpeakerScraperJobProcessor.js
slifty: No.

reefdog: Have it your way dan.

src/server/queues/knownSpeakerScraperQueue/knownSpeakerScraperJobProcessor.js
slifty: I wonder if we should change `claimBusterClaimDetectorJobProcessor` as well.  I will consider making an issue.

src/server/queues/knownSpeakerScraperQueue/knownSpeakerScraperJobProcessor.js
slifty: Look don't worry you'll see the change eventually.

src/server/queues/knownSpeakerScraperQueue/knownSpeakerScraperJobProcessor.js
reefdog: I'm gonna create an `optimization` label you can apply to that if you do. (Also applies to #252.)

slifty: Can we actually have that label be `premature-optimization`

reefdog: Yes but paradoxically we'll never apply it because creating the issue for later is the actual solution to premature optimizing.

src/server/queues/knownSpeakerScraperQueue/knownSpeakerScraperJobProcessor.js
```diff
@@ -0,0 +1,47 @@
...
+  }, transaction)
+
+const saveKnownSpeakers = (knownSpeakers, transaction) => KnownSpeaker.bulkCreate(
+  knownSpeakers,
+  transaction,
+)
+
+export default async () => {
+  const documentId = config.KNOWN_SPEAKERS_GOOGLE_DOC_ID
+  logger.info(`Syncing known speakers from GOOGLE doc: ${documentId}`)
```
slifty: Glad that Google is finally capitalized

PR #291, opened 2020-01-15
Update Claimbuster API endpoint

### LINKED ISSUE #290: Switch to new production ClaimBuster API endpoint
We're currently [pointing to the dev version of the current ClaimBuster API](https://github.com/TechAndCheck/tech-and-check-alerts/blob/eff2df9f9e25565b2d28a7074ccbf4c56816acb2/src/server/constants.js#L24), which is all we've had.

They recently added a production-stable version, at `https://idir.uta.edu/claimbuster/api/v1` (note the lacking `-dev`). We should change!

reefdog:
Previously, Claimbuster only offered the dev version of the API.

They recently added a production version and requested we switch to it.

Resolves #290

### DIFF
--- a/src/server/constants.js
+++ b/src/server/constants.js
@@ -21,4 +21,4 @@ export const TWITTER_LIST_NAMES = {
 
 export const CLAIMBUSTER_THRESHHOLD = 0.5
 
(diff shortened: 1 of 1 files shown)


reefdog:
little-larry-sellers

PR #299, opened 2020-01-18
Repair newsletter display of claim speaker

### LINKED ISSUE #297: Speaker names not appearing in newsletters
This morning's newsletters are missing speaker names:

Screen Shot 2020-01-18 at 9 26 22 AM

This is because in #294 we corrected the Claims model but forgot to update the newsletter templates accordingly. (It also demonstrates a failure of process: this wasn't caught in review and while we tested the scraper — which is what we'd set out to fix — we didn't test the final newsletter content. Further need for [full integration testing](https://github.com/TechAndCheck/tech-and-check-alerts/issues/227).)

reefdog:
In #268 and #294, we updated the Claim model to split `speakerName` into two fields, (`speakerExtractedName` and `speakerNormalizedName`, with the latter intended for display) but didn’t flow this change down to the newsletter.

This change correctly displays the normalized name in the newsletter.

Resolves #297

### DIFF
--- a/src/server/newsletters/templates/national.hbs
+++ b/src/server/newsletters/templates/national.hbs
@@ -10,7 +10,7 @@
         <div class="claim__metadata">
           <a href="{{canonicalUrl}}" class="claim__permalink">
             <cite class="claim__speaker">
-              {{speakerName}}
+              {{speakerNormalizedName}}
             </cite>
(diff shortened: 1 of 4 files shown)


reefdog:
little-larry-sellers

Issue #317, opened 2020-02-06
Add support for a test database


slifty:
As we move to the land of integration (#227) and component tests (#309) we need to set up our test suite to support interaction with a test database.  This will apply both for local development as well as CI and deployment, since tests are run as part of both CI and CD.

There are a few approaches here, which ultimately come down to "when do we run migrations."

I suggest we start off by [running migrations once](https://medium.com/riipen-engineering/testing-with-sequelize-cc51dafdfcf4), as part of the [`pretest`](https://docs.npmjs.com/misc/scripts) command.

Another option would be building and dropping the database [before every test](https://jestjs.io/docs/en/setup-teardown) or manually in tests that might interact with the database.  I believe this would be very slow / overkill.  Instead, tests that care about database state in order to run should set that state directly.

## Necessary steps

Regardless of final decision, the following will have to happen before tests can invoke code that call the database.

1. Test database migrations get run.
2. Database connections closed at the end of the suite.

## Other considerations

As part of this issue we may need to remove (or modify) the `test:ci` command since it does not automatically call the `pretest` and `posttest` hooks.  I'm leaning towards removing the idea of `test:ci` and `test:watch` in favor of a single `yarn test` command that can be modified with flags as desired.

slifty:
In pursuing this issue I'm starting to think that in addition to support for a test database we will also want to support database mocks.  The mocks would be for unit tests that happen to engage with a database, where database connections are unnecessary and potentially bulky.  The test database would be for integration tests, where the output of a given component is going to fuel another component and where data needs to be queried in more advanced ways.

I'll make a separate issue for mocks.

PR #321, opened 2020-02-12
Add support for a test database

slifty:
This adds in the pipework to allow our tests to interact with a database.

It migrates the test database when tests are run, adds Travis support for postgres, adds some protection to make sure tests are never run on a production / dev database, and also fixes an issue with database connections that became a blocker after enabling the test DB on CI.

This does not look like a week of work but there was a big struggle / rabbit hole as outlined in [this SO issue](https://stackoverflow.com/questions/60217417/jest-tests-hang-due-to-open-sequelize-connections)

## To test this PR

Just run tests. (`yarn test`)

## Extra Setup

You will probably find that you need to update your environment variables so that your test DB is different from your prod / dev DB.  Since we weren't using test DB in the past it's likely that they were all pointing to the same place.

Resolves Issue #317

### DIFF
--- a/config/sequelize-config.js
+++ b/config/sequelize-config.js
@@ -1,6 +1,31 @@
+const assert = require('assert')
 const dotenv = require('dotenv')
 
 dotenv.config()
+
+// I'm not sure if this should be in an external utility, but given how odd sequelize config
+// is I would like to have this check be directly next to the area where configuration
+// is defined, in case it ever changes.
+//
+// It is CRITICAL that the test database never match produciton / dev, since tests can be
+// destructive to data.
+//
+// I'm wrapping this check in a method name anyway just for the sake of readability.
+const assertSafeDatabaseConfiguration = function() {
+  const testDatabaseURL = new URL(process.env.DATABASE_URL_TEST)
+  const devDatabaseURL = new URL(process.env.DATABASE_URL_DEVELOPMENT)
+  const prodDatabaseURL = new URL(process.env.DATABASE_URL_PRODUCTION)
+  assert(
+    testDatabaseURL.pathname !== devDatabaseURL.pathname,
+    `Test DB (${testDatabaseURL.pathname}) cannot match Development DB (${devDatabaseURL.pathname})`,
+  )
+  assert(
+    testDatabaseURL.pathname !== prodDatabaseURL.pathname,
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "start": "concurrently 'yarn start-node' 'yarn parcel:dev'",
     "build": "babel src -d lib",
     "lint": "./node_modules/.bin/eslint 'src/**/*.js'",
+    "pretest": "NODE_ENV=test yarn migrate",
     "test": "jest",
     "test:watch": "jest --verbose --watch",
```
reefdog: Do we know if `yarn test:watch` will still fire `pretest`?

slifty: I do not believe it will.

reefdog: We should verify that and kill it too then, yeah?

slifty: It will still run in some circumstances (though you will have to have run test once to ensure migrations happened, or manually run pretest).  But yes I think that it is just a helper that doesn't need to exist any longer.

If it is useful as syntactic sugar, we could replace "jest --verbose --watch" with "yarn test --verbose --watch"

reefdog: It's theoretically useful, but I never remember it's there.

reefdog: But I'd say let's pick one of those: kill it or ensure it acts as we expect (running `pretest`).

config/sequelize-config.js
```diff
@@ -1,7 +1,32 @@
...
 
+// I'm not sure if this should be in an external utility, but given how odd sequelize config
+// is I would like to have this check be directly next to the area where configuration
+// is defined, in case it ever changes.
+//
+// It is CRITICAL that the test database never match produciton / dev, since tests can be
+// destructive to data.
+//
+// I'm wrapping this check in a method name anyway just for the sake of readability.
+const assertSafeDatabaseConfiguration = function() {
```
reefdog: Oops, missed something: this should be ES6-style, right?

`const assertSafeDatabaseConfiguration = () => {`

slifty: ... IT CHANGED!

This file is loaded directly by sequelize, and in the past I swear it errored out when using ES6 (which is why this file is not linted).  Specifically import vs require.

It looks like it all works now though, and a test change to import earlier in the file caused no issues.

Right now the file is half ES5 (require) and half ES6 (use of const)...
Would it be OK if I make an issue to fix the whole file at once and merge this as is?

reefdog: Yes for sure!

PR #324, opened 2020-02-14
Remove test:ci command

slifty:
When we set up CI we added a `test:ci` command.  This was because our normal test command was set to use --verbose, and we didn't need all that noise in our CI.

The issue is that `test` is actually a special command, which will invoke `pretest` automatically.  We want our CI to benefit from these hooks.

The other problem, by the way, is that --verbose is super noisy everywhere including on dev and so we actually shouldn't be forcing it to be flagged every time we want to run `yarn test` anyway.

This removes `test:ci` and makes `test` less noisy.

Issue #317


slifty:
I am approving this on behalf of @reefdog -- I didn't actually give him a chance to respond but I

slifty:
Justin did weigh in and give a 👍 to this PR.  One comment was around whether we should remove the coverage and yes I felt like ultimately we should.  That said, we may want to consider working coverage reporting into our flow somehow.

Until we do that, I feel like it was just kind of noise.

PR #327, opened 2020-02-17
Remove ssh_known_hosts TravisCI addon

### LINKED ISSUE #322: Travis accesses env variables before setting them
Looking at the travis logs I noticed:

```
Adding ssh known hosts

0.06s$ ssh-keyscan -t $TRAVIS_SSH_KEY_TYPES -H \$STAGING_SERVER 2>&1 | tee -a ${TRAVIS_HOME}/.ssh/known_hosts

getaddrinfo $STAGING_SERVER: Name or service not known

getaddrinfo $STAGING_SERVER: Name or service not known

getaddrinfo $STAGING_SERVER: Name or service not known
```

It looks as if the:

```
addons:
  ssh_known_hosts:
  - "$STAGING_SERVER"
```

of travis.yml might not be meaningful / might be run before $STAGING_SERVER is populated with the configured value.

Lets either remove that section of the yml, or fix it so it doesn't generate log noise.

slifty:
TravisCI does not allow the use of environment variables when setting known hosts.  This has two effects for us:

1. It means that the ssh_known_hosts apparently never did anything
2. It means TravisCI has some error output when trying to set up known
   hosts using environment vars.

This removes the offending attempt.  I don't expect there to be any consequence for our CD processes.

Resolves #322

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 language: node_js
 
-addons:
-  ssh_known_hosts:
-  - "$STAGING_SERVER"
-
 before_script:
   - echo "Host $STAGING_SERVER" >> ~/.ssh/config
   - echo "StrictHostKeyChecking no" >> ~/.ssh/config


slifty:
As part of continuous deployment we use SSH to deliver the project files to dev (and eventually prod) and then SSH again to run a script on that server to restart the process.

Adding something to your known hosts file prevents that message you get from the ssh client saying something like "sup dog never seen this host before are you sure it's OK?" which usually req-- ok my two minutes are up

slifty:
uires a response.

This is already avoided, though, through this line in travis.yml:

```
  - echo "StrictHostKeyChecking no" >> ~/.ssh/config
```

reefdog:
Got it! Yeah, I figured that was the intent, just wasn't sure how we were getting around the need for it. Your wantonly excessive second comment makes that clear. FULLY APPROVED.

PR #331, opened 2020-02-25
Store scrape log data more efficiently

### LINKED ISSUE #137: Stop storing results and errors in the scrape logs
It was a fine idea at the time but the reality is that storing scrape results and errors is a LOT of space and if we really need it for debugging we can always re-run the scrape in our sandbox and play with the outcomes then.

The database balloons right now, but this will be corrected with #136, making this particular issue less urgent.

We still want to store the outcome of the scrape (success vs failure), likely just as a boolean.

reefdog:
## Description

Previously, we stored each scrape's complete `result` or `error` response data in the `scrape_log` table for every scrape logged. This was fine during development, but is much weightier than necessary for permanence. For instance, we've accumulated 4.6GB of `result` data in production in fewer than five months. (We've stored 5MB of `error` data in the same time.)

This change replaces the `scrape_log.result` and `scrape_log.error` text columns with two new columns:

- `scrape_log.scrape_response_code` for storing the HTTP response code, or a `non-http-error` fallback
- `scrape_log.scrape_response_message` for storing an optional longer message, currently only used for storing error messages

It also adapts the `ScrapeLog` model to use these columns and includes a complex migration that coalesces legacy data into the new fields (and back, upon migration reversal).

However! This change intentionally removes the accumulated `scrape_log.result` data, so reversing the migration will _not_ restore previous successful result data.

**🚨Let me reiterate this very important point:🚨** This change is **destructive** and running the migration will **delete the `scrape_logs.result` data**. Undoing the migration will restore error data (since we keep those messages), but the `result` column for successful scrapes will simply contain the string `"200"`. (Yes, the migration survives being run and reversed repeatedly, with the caveat that original `result` contents will be gone [forever](https://getyarn.io/yarn-clip/172b7a36-0c61-49c5-beea-e3cab6065066).)

## Steps to Test

1. Look at your database and find both successful and unsuccessful `scrape_logs` rows (presence of `result` or `error` content).
2. Run `yarn migrate`.
3. Check those same rows; they should now have a `scrape_response_code` column set to `"200"` if they were successful or their appropriate HTTP error code (or `"non-http-error"`) if not, and errors should be stored in `scrape_response_message`.
4. Fire up the scraper and let it scrape some Twitter or CNN content.
5. Make sure new `scrape_logs` rows were created with expected response values.
6. To be real fancy, reverse the migration (`sequelize db:migrate:undo`) and make sure `result`/`error` columns are restored (sans original `result` data); just run `yarn migrate` again when you're done being impressed with me.

## Deploy Notes

- After deploying we will need to migrate with `yarn migrate`

## Related Issues

Resolves #137

### DIFF
--- a/src/server/migrations/20200225232115-refactor-scrape-logs-response-status.js
+++ b/src/server/migrations/20200225232115-refactor-scrape-logs-response-status.js
@@ -1,0 +1,97 @@
+import { runPromiseSequence } from '../utils'
+
+import { SCRAPE_RESPONSE_CODES } from '../workers/scrapers/constants'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => queryInterface.sequelize
+    .transaction(t => runPromiseSequence([
+      () => queryInterface.addColumn(
+        'scrape_logs',
+        'scrape_response_code',
+        {
+          type: Sequelize.STRING,
+        },
+        { transaction: t },
+      ),
+      () => queryInterface.addColumn(
+        'scrape_logs',
+        'scrape_response_message',
+        {
+          type: Sequelize.TEXT,
+        },
+        { transaction: t },
+      ),
+      () => queryInterface.sequelize.query(
+        `
+        UPDATE scrape_logs
+        SET    scrape_response_code = '${SCRAPE_RESPONSE_CODES.HTTP_SUCCESS}'
+        WHERE  result IS NOT NULL`,
+        { transaction: t },
+      ),
+      () => queryInterface.sequelize.query(
+        `
+        UPDATE scrape_logs
+        SET    scrape_response_code = COALESCE(
+                 SUBSTRING(error from '^[0-9]{3}'),
+                 '${SCRAPE_RESPONSE_CODES.NON_HTTP_ERROR}'
+                ),
+               scrape_response_message = error
+        WHERE  error IS NOT NULL`,
+        { transaction: t },
+      ),
+      () => queryInterface.removeColumn(
+        'scrape_logs',
+        'result',
+        { transaction: t },
+      ),
+      () => queryInterface.removeColumn(
+        'scrape_logs',
+        'error',
+        { transaction: t },
+      ),
+    ])),
+  down: (queryInterface, Sequelize) => queryInterface.sequelize
+    .transaction(t => runPromiseSequence([
+      () => queryInterface.addColumn(
+        'scrape_logs',
+        'result',
+        {
+          type: Sequelize.TEXT,
+        },
+        { transaction: t },
+      ),
+      () => queryInterface.addColumn(
+        'scrape_logs',
+        'error',
+        {
+          type: Sequelize.TEXT,
+        },
+        { transaction: t },
+      ),
+      () => queryInterface.sequelize.query(
+        `
+        UPDATE scrape_logs
+        SET    result = '${SCRAPE_RESPONSE_CODES.HTTP_SUCCESS}'
+        WHERE  scrape_response_code = '${SCRAPE_RESPONSE_CODES.HTTP_SUCCESS}'`,
+        { transaction: t },
+      ),
+      () => queryInterface.sequelize.query(
+        `
+        UPDATE scrape_logs
+        SET    error = scrape_response_message
+        WHERE  scrape_response_code != '${SCRAPE_RESPONSE_CODES.HTTP_SUCCESS}' OR
+               scrape_response_code IS NULL`,
+        { transaction: t },
+      ),
+      () => queryInterface.removeColumn(
+        'scrape_logs',
+        'scrape_response_code',
+        { transaction: t },
+      ),
+      () => queryInterface.removeColumn(
+        'scrape_logs',
+        'scrape_response_message',
+        { transaction: t },
+      ),
+    ])),
+}
--- a/src/server/models/ScrapeLog.js
+++ b/src/server/models/ScrapeLog.js
@@ -2,8 +2,8 @@ module.exports = (sequelize, DataTypes) => {
   const ScrapeLog = sequelize.define('ScrapeLog', {
     scrapeUrl: DataTypes.STRING(1024),
     scraperName: DataTypes.STRING,
-    result: DataTypes.TEXT,
-    error: DataTypes.TEXT,
+    scrapeResponseCode: DataTypes.STRING,
+    scrapeResponseMessage: DataTypes.TEXT,
   }, {})
   return ScrapeLog
 }
--- a/src/server/workers/scrapers/AbstractScraper.js
+++ b/src/server/workers/scrapers/AbstractScraper.js
@@ -3,6 +3,7 @@ import dayjs from 'dayjs'
 
 import logger from '../../utils/logger'
 import models from '../../models'
+import { SCRAPE_RESPONSE_CODES } from './constants'
 
 const { ScrapeLog } = models
 
@@ -125,7 +126,7 @@ class AbstractScraper {
       where: {
         scrapeUrl: this.getScrapeUrl(),
         scraperName: this.getScraperName(),
-        error: null,
+        scrapeResponseCode: SCRAPE_RESPONSE_CODES.HTTP_SUCCESS,
       },
       order: [['createdAt', 'DESC']],
     })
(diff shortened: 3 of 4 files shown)


src/server/models/ScrapeLog.js
```diff
@@ -2,8 +2,7 @@ module.exports = (sequelize, DataTypes) => {
   const ScrapeLog = sequelize.define('ScrapeLog', {
     scrapeUrl: DataTypes.STRING(1024),
     scraperName: DataTypes.STRING,
-    result: DataTypes.TEXT,
-    error: DataTypes.TEXT,
+    wasSuccessful: DataTypes.BOOLEAN,
```
reefdog: @slifty I fully expect you to 🤢 at this column name.

src/server/migrations/20200225232115-add-was-successful-to-scrape-logs.js
```diff
@@ -0,0 +1,72 @@
...
+        { transaction: t },
+      ),
+      () => queryInterface.sequelize.query(
+        `
+        UPDATE scrape_logs
+        SET    was_successful = TRUE
+        WHERE  result IS NOT NULL`,
+        { transaction: t },
+      ),
+      () => queryInterface.removeColumn(
```
slifty: Since this migration is called "add_____" I have this instinct that we should put the removes in their own migration(...s?).  That way the migration(s!) are a bit more atomic.

Alternatively maybe a rename of the migration to something like `refactor-scrape-logs-success-state` which indicates a more complex situation.

src/server/models/ScrapeLog.js
slifty: uh oh I didn't mind it... though I guess it's missing a noun (`wasSuccessfulScrape` or `wasSuccessfulOutcome` or `wasSuccessfulAtMakingChocolateFromScratch`)

src/server/workers/scrapers/AbstractScraper.js
```diff
@@ -137,15 +137,13 @@ class AbstractScraper {
   /**
    * Generates and saves a scrape log entry
    *
-   * @param  {String} result The stringified result of the scrape
-   * @param  {String} error  The error message, if any
+   * @param  {Boolean} wasSuccessful True if the scrape was successful, else false
    * @return {ScrapeLog} the unsaved initial scrape log entry
    */
-  createScrapeLog = (result, error) => ScrapeLog.create({
+  createScrapeLog = wasSuccessful => ScrapeLog.create({
```
slifty: Oh I do hate it here though.

`wasGreatSuccess`
`successAchievementUnlocked`
`willHaveBeenSuccess`
`wasntSuccessful`

what about instead of a success boolean we make this an outcome enum instead of a boolean, and for now it is just "SUCCESS" or "FAILURE" but maybe some day it can expand to be "STATUS404" 

Or maybe it should be `statusCode` since scrapes will always be HTTP.  Actually wait yeah shouldn't we do that?  status code is way more interesting than success boolean.

This was good!  (`appleInc`)

src/server/workers/scrapers/AbstractScraper.js
```diff
@@ -163,13 +161,13 @@ class AbstractScraper {
         logger.debug(`Success (${this.scrapeUrl})`)
         this.setScrapeResponse(responseString)
         const result = await this.scrapeHandler(responseString)
-        this.createScrapeLog(JSON.stringify(result))
+        this.createScrapeLog(true)
```
slifty: I know I did this but I'm realizing that maybe we should be passing an object so it's super clear what the attributes mean.  E.g.

```
this.createScrapeLog({
  appleInc: result.status,
})
```

src/server/migrations/20200225232115-add-was-successful-to-scrape-logs.js
reefdog: You are correct and IMO the fix should be renaming the migration. (You may want to remember to `sequelize db:migrate:undo` _before_ pulling again, unless we want to have faith that Sequelize won't be confused by a migration file name change.)

slifty: Yes of course, before pulling AGAIN!

src/server/workers/scrapers/AbstractScraper.js
reefdog: TRUE STORY: This was my desire too and I even [read up on the best way to start down this pattern in JS](https://medium.com/dailyjs/named-and-optional-arguments-in-javascript-using-es6-destructuring-292a683d5b4e) but assumed you'd push back against introducing a new pattern we weren't using elsewhere. If you're cool with it, I vastly approve of used named arguments when reasonable.

src/server/workers/scrapers/AbstractScraper.js
reefdog: I really like this.

src/server/migrations/20200225232115-refactor-scrape-logs-response-status.js
```diff
@@ -0,0 +1,107 @@
...
+        'scrape_logs',
+        'result',
+        { transaction: t },
+      ),
+      () => queryInterface.removeColumn(
+        'scrape_logs',
+        'error',
+        { transaction: t },
+      ),
+      () => queryInterface.addIndex(
```
slifty: This index should be a pair of fields -- scrape_response_code and scrape_type

Only one index is used in a given query, so we should use both of the fields that our lookup query is going to be using (we could create two indexes if we think we'll ever be querying just in terms of response code)

src/server/migrations/20200225232115-refactor-scrape-logs-response-status.js
```diff
@@ -0,0 +1,107 @@
...
+        'scrape_logs',
+        'scrape_response_code',
+        { transaction: t },
+      ),
+      () => queryInterface.removeColumn(
+        'scrape_logs',
+        'scrape_response_message',
+        { transaction: t },
+      ),
+      () => queryInterface.removeIndex(
```
slifty: Huh; do you have to remove the index if you already removed the column?

src/server/workers/scrapers/constants.js
```diff
@@ -22,3 +22,8 @@ export const STATEMENT_SCRAPER_PLATFORMS = {
   CNN_TRANSCRIPT: 'CNN',
   TWITTER_ACCOUNT: 'TWITTER',
 }
+
+export const SCRAPE_RESPONSE_CODES = {
+  NON_HTTP_ERROR: 'non-http-error',
+  SUCCESS: '200',
```
slifty: should this be HTTP_SUCCESS just to be explicit that we're mapping to HTTP codes?

src/server/migrations/20200225232115-refactor-scrape-logs-response-status.js
reefdog: You're right, you don't!

> When you remove a column from a table, PostgreSQL will automatically remove all of its indexes and constraints involving the column.

I'll remove.

src/server/workers/scrapers/constants.js
reefdog: Yes!

reefdog: Although instead of what you suggested I think I'm going to go with `HTTP_SUCCESS`, monospaced instead of sans-serif the way you did it. So see it's an original idea.

src/server/migrations/20200225232115-refactor-scrape-logs-response-status.js
reefdog: By `scrape_type` I assume you mean `scraper_name`?

Also, the actual query uses three fields (`scraper_name`, `scrape_url`, and `scrape_response_code`). Should the index by on all three? Seems heavy but I'm also not an indexpert.

slifty: Right!  Sure why not!

I believe order will matter...  Maybe go most broad to most specific?  name -> url -> response

reefdog: Actually… we already have an index on scraper name + URL. I'm gonna call that enough for the major performance implications and remove this index entirely.

slifty:
Ready for re-re-re-@reefdog-review

Issue #255, opened 2019-09-17
Add ability to deactivate Twitter accounts


reefdog:
Right now, all Twitter accounts in the database are scraped. It would be nice to be able to deactivate accounts temporarily without purging them entirely from the database. (For instance, we've discovered a handful of accounts are 404ing and 401ing, so we'd like to deactivate scraping them until we decide what to do.)

As a side note: this affects #229, because the fastest way to sync an external master Google Sheet is just to delete anything that doesn't exist, update or create anything that does. We'll need to replace deleting with deactivating, and add the concept of reactivating.

Additional note (also being left on #229): we have to assume the same account could be in more than one list, so we should be sure to only take action on Twitter accounts based on their status in the current list.

reefdog:
#355 resolved this woo, and the behavior referenced in the description is captured by #216.

PR #355, opened 2020-04-17
Add twitter account sync

### LINKED ISSUE #229: Sync Twitter account lists from Google Sheet

slifty:
## Description
This PR adds two new queues in order to create the regular synchronization of twitter accounts with a google doc.

The first queue will scrape the document associated with a list, disable twitter accounts associated with that list, and re-enable (or add) any accounts that are on the list according to the scrape.

The second queue is the initiation queue which starts jobs in the first queue for each registered twitter list.

A critical note: this PR adds calculated tech debt.  Specifically, we are storing the google docs as constants.  This is NOT an appropriate use of a constant, but since we are about to change the way twitter lists and newsletters are managed in our system we felt that this was the least onerous approach.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits
- [x] Relevant labels assigned

## Steps to Test
1. `yarn test`

## Deploy Notes
- `yarn migrate`

## Related Issues
Resolves #229 
Related to #255

### DIFF
--- a/src/server/constants.js
+++ b/src/server/constants.js
@@ -19,6 +19,15 @@ export const TWITTER_LIST_NAMES = {
   NORTH_CAROLINA: 'northCarolina',
 }
 
+// TODO - This should NOT be a constant. It should be DB driven. We are
+// using a constant temporarily since the dynamic newsletter functionality
+// will potentially change the way twitter lists are organized, and didn't
+// want to create migrations that will be almost immediately obsolete.
+export const TWITTER_LIST_GOOGLE_DOC_ID_MAP = {
+  [TWITTER_LIST_NAMES.NATIONAL]: '1gLkx2LK3yhS-glpsktWYNqBc9H2zsd7C6TvmgWjL5Kg',
+  [TWITTER_LIST_NAMES.NORTH_CAROLINA]: '1Z7grw_GQLNMtSVvkUtbXatly_JId7h3yVDX9BWyJetg',
+}
+
 export const CLAIMBUSTER_THRESHHOLD = 0.5
 
 export const CLAIMBUSTER_API_ROOT_URL = 'https://idir.uta.edu/claimbuster/api/v1'
--- a/src/server/models/TwitterAccount.js
+++ b/src/server/models/TwitterAccount.js
@@ -2,9 +2,15 @@ const Sequelize = require('sequelize')
 
 module.exports = (sequelize, DataTypes) => {
   const TwitterAccount = sequelize.define('TwitterAccount', {
-    screenName: DataTypes.STRING(128),
+    screenName: {
+      type: DataTypes.STRING(128),
+      unique: 'twitterAccountScreenNameListName',
+    },
     preferredDisplayName: DataTypes.STRING(512),
-    listName: DataTypes.STRING(128),
+    listName: {
+      type: DataTypes.STRING(128),
+      unique: 'twitterAccountScreenNameListName',
+    },
     isActive: DataTypes.BOOLEAN,
   }, {})
 
--- a/src/server/models/TwitterAccount.js
+++ b/src/server/models/TwitterAccount.js
@@ -45,5 +45,26 @@ module.exports = (sequelize, DataTypes) => {
   TwitterAccount.getActiveByListName = async listName => TwitterAccount
     .getActiveByListNames([listName])
 
+  TwitterAccount.deactivateTwitterAccountsByList = async (listName, transaction) => TwitterAccount
+    .update(
+      { isActive: false },
+      {
+        where: {
+          listName,
+        },
+        transaction,
+      },
+    )
+
+  TwitterAccount.createOrActivateTwitterAccounts = async (twitterAccounts, transaction) => Promise
+    .all(
+      twitterAccounts.map(
+        account => TwitterAccount.upsert(
+          account,
+          { transaction },
+        ),
+      ),
+    )
+
   return TwitterAccount
 }
--- a/src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
+++ b/src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
@@ -1,0 +1,78 @@
+import GoogleSpreadsheetScraper from '../../workers/scrapers/GoogleSpreadsheetScraper'
+import logger from '../../utils/logger'
+import models from '../../models'
+import {
+  decorateObjects,
+  hasKey,
+} from '../../utils'
+import {
+  extractScreenName,
+  isTwitterScreenName,
+} from '../../utils/twitter'
+import { getDocumentIdByTwitterListName } from '../../utils/google'
+
+const {
+  sequelize,
+  TwitterAccount,
+} = models
+
+const isTwitterDisplayNameColumn = columnName => /twitter_\d+/.test(columnName)
+  || columnName === 'twitter_%d'
+
+const extractTwitterScreenNameColumnsFromRow = row => Object.keys(row)
+  .filter(isTwitterDisplayNameColumn)
+
+const hasAtLeastOneTwitterScreenNameColumn = row => (
+  extractTwitterScreenNameColumnsFromRow(row).length >= 1
+)
+
+const isValidTwitterAccountRow = row => (
+  hasKey(row, 'display_name')
+  && hasAtLeastOneTwitterScreenNameColumn(row)
+)
+
+const extractTwitterAccountsFromRow = (row, screenNameColumns) => {
+  const screenNames = screenNameColumns
+    .map(columnName => extractScreenName(row[columnName]))
+    .filter(isTwitterScreenName)
+  const twitterAccounts = screenNames.map(screenName => ({
+    screenName,
+    preferredDisplayName: row.display_name,
+  }))
+  return twitterAccounts
+}
+
+const extractTwitterAccountsFromRows = (rows, screenNameColumns) => rows
+  .filter(isValidTwitterAccountRow)
+  .map(row => extractTwitterAccountsFromRow(row, screenNameColumns))
+  .reduce(
+    (accumulator, twitterAccounts) => accumulator.concat(twitterAccounts),
+    [],
+  )
+
+export default async (job) => {
+  const {
+    data: {
+      listName,
+    },
+  } = job
+
+  const documentId = getDocumentIdByTwitterListName(listName)
+  logger.info(`Syncing twitter accounts for "${listName}" from Google doc: ${documentId}`)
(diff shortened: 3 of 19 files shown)


src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
```diff
@@ -0,0 +1,53 @@
...
+  hasKey(row, 'display_name')
+  && hasKey(row, 'twitter_1')
+)
+
+const extractTwitterAccountsFromRows = rows => rows
+  .filter(isValidTwitterAccountRow)
+  .map(row => ({
+    screenName: extractScreenName(row.twitter_1),
+    preferredDisplayName: row.display_name,
+  }))
```
reefdog, line 26: From the hard-coded `twitter_1`, it looks like we aren't ever scraping `twitter_2` (or further). Shouldn't we be?

src/server/models/TwitterAccount.js
```diff
@@ -39,5 +45,26 @@ module.exports = (sequelize, DataTypes) => {
...
+
+  TwitterAccount.createOrActivateTwitterAccounts = async (twitterAccounts, transaction) => Promise
+    .all(
+      twitterAccounts.map(
+        account => TwitterAccount.upsert(
+          account,
+          { transaction },
+        ),
+      ),
+    )
```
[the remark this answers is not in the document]
slifty: Unfortunately no(t yet) -- my first implementation used bulkCreate and updateOnDuplicate but it [has a bug where unique indexes that involve multiple columns does not work](https://github.com/sequelize/sequelize/issues/11569).

That bug is fixed, but the patch isn't released.  We could make a tracking issue and make the optimization at a later date when that bug fix is released in Sequelize though!

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
slifty: Right; I meant to make this smarter so that it would detect any row starting named twitter_%d.  I'll add that in.

src/server/constants.js
```diff
@@ -19,6 +19,15 @@ export const TWITTER_LIST_NAMES = {
   NORTH_CAROLINA: 'northCarolina',
 }
 
+// TODO - This should NOT be a constant. It should be DB driven. We are
+// using a constant temporarily since the dynamic newsletter functionality
+// will potentially change the way twitter lists are organized, and didn't
+// want to create migrations that will be almost immediately obsolete.
+export const TWITTER_LIST_DOCUMENT_MAP = {
```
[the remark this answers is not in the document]
slifty: Good suggestion ty!

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
reefdog: Sounds great except make sure that it's `twitter_` followed by a number, not followed by `%d`.

↑ see dan this is how you troll

src/server/models/TwitterAccount.js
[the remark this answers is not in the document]
reefdog: I'm just happy to know we share priorities!

src/server/constants.js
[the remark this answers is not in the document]
reefdog: Please be sure to give me co-author credit on all commits that use that variable name in perpetuity thank you. People are talking about this Justin Reese more and more.

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
slifty: I don't get it

reefdog: YES YOU DO

src/server/constants.js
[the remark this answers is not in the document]
slifty: Done but sorry I don't rewrite history so I couldn't add you as a co author

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
slifty: The ultimate troll was the amount of time you made me spend implementing this.

slifty:
This is ready for re-review @reefdog

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
```diff
@@ -0,0 +1,74 @@
...
+  isTwitterScreenName,
+} from '../../utils/twitter'
+import { getDocumentIdByTwitterListName } from '../../utils/google'
+
+const {
+  sequelize,
+  TwitterAccount,
+} = models
+
+const isValidTwitterAccountRow = row => (
```
reefdog, line 19: It would be nice if this function was documented to explain why those two specific checks result in a valid twitter account row, though maybe my initial confusion as to why `twitter_1` was still hard-coded was purely due to past knowledge of the `twitter_1`-hard-coding problem. (I get it now: we need _at least_ a display name and a twitter screen name, but that took a bit of staring, and could be improved with a simple "We need at least…" comment.)

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
```diff
@@ -0,0 +1,74 @@
...
+  hasKey(row, 'display_name')
+  && hasKey(row, 'twitter_1')
+)
+
+const isTwitterDisplayNameColumn = columnName => /twitter_\d+/.test(columnName)
+  || columnName === 'twitter_%d'
+
+const extractTwitterScreenNameColumnsFromRow = row => Object.getOwnPropertyNames(row)
```
reefdog, line 27: Out of curiosity, what's the practical difference between `Object.getOwnPropertyNames` and `Object.keys`? I think we've used `Object.keys` more often in the past.

src/server/utils/__test__/twitter.test.js
```diff
@@ -94,6 +95,16 @@ describe('getSpeakerAffiliationFromTweet', () => {
...
   })
 })
+describe('extractScreenName', () => {
+  it('Should extract screen names', () => {
+    expect(extractScreenName('https://www.twitter.com/SenatorBurr')).toEqual('SenatorBurr')
+    expect(extractScreenName('http://wwww.twitter.com/jeffnc2')).toEqual('jeffnc2')
+    expect(extractScreenName('http://twitter.com/CalforNC')).toEqual('CalforNC')
+    expect(extractScreenName('http://twitter.com/CalforNC?q=15')).toEqual('CalforNC')
+    expect(extractScreenName('@slifty')).toEqual('slifty')
+    expect(extractScreenName('@slifty_with_underscores')).toEqual('slifty_with_underscores')
```
reefdog, line 105: We'll need to expand these tests to correctly return nothing when handed a URL for a non-Twitter domain, such as `https://www.example.com/foo`.

src/server/utils/twitter.js
```diff
@@ -42,6 +42,19 @@ export const extractStatementsFromTweets = tweets => tweets.map(tweet => ({
...
+ *
+ * The string could be a URL, or an @handle, or just a handle.
+ *
+ * @param  {String} str The string containing a twitter screen name.
+ * @return {String}     The extracted screen name, hopefully
+ */
+export const extractScreenName = str => str
+  .replace(/http(s)?:\/\/(.*\.)?twitter\.com\//, '')
+  .replace('@', '')
+  .replace(/([a-zA-Z0-9_]+).*/, '$1')
```
reefdog, line 56: Unfortunately, dirty source data means we should reject stuff that doesn't look like a screen name or a Twitter-based URL. `extractScreenName('https://example.org/foo')` currently returns `https`, when it should return nothing.

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
slifty: The main difference is that Object.getOwnPropertyNames had a higher SEO with the specific terms I used when looking it up this time.

The real difference is [interesting](https://stackoverflow.com/questions/22658488/object-getownpropertynames-vs-object-keys) -- we should be using Object.keys (though it will not change behavior).

src/server/utils/twitter.js
slifty: Overhauling this method -- The new version will also have the added benefit of never returning an invalid screen name.

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
slifty: I guess technically we don't even *need* that; nothing will break if it doesn't exist...

I could just take that requirement out completely, what do you think?

reefdog: Do it!

slifty: OK ANOTHER THOUGHT

This could be:

```

const isValidTwitterAccountRow = row => (
  hasKey(row, 'display_name')
  && extractTwitterScreenNameColumnsFromRow(row).length >= 1
)
```

The tradeoff is (minimal) computation time vs self-documented expectations around what a "valid" twitter account row is.

e.g. if there is no twitter handle column, it won't *break* but is it *valid*?

slifty: (ok that's what I did; feels better.)

reefdog: Okay!

slifty:
fixed again!

PR #365, opened 2020-06-11
Move Twitter lists to the database

### LINKED ISSUE #364: Make twitters lists DB driven
In the effort towards #360 we need the twitter lists to be a database entity.  This means:

1. Creating a new "twitter list" table.
2. Removing the concept of "list names" and instead using primary keys.

* Adding a foreign key relationship between twitter accounts and twitter lists
* Changing twitter sync to load lists / list data from DB rather than constants

This might also mean changing the newsletter generation, but we can be a little more ugly there since that is being actively rewritten (and indeed this issue is just an attempt to split up the PRs associated with that rewrite).

### LINKED ISSUE #368: Change how the test database is constructed
Right now we "tear down" and "build up" the test database whenever `yarn test` is run.  This is currently done via `db:migrate:undo:all` which has the benefit of testing migration teardowns, but the cost of causing very pesky errors whenever you change branches.

The issue is that one branch might have a migration (that is then in the test db) which is not in another.  This means that when you run `yarn test` the second time, it fails because it can't find the original migration in order to roll it back.

Sequelize has `db:drop` and `db:create` which I had previously assumed were not available for us in postgres, but recently learned [actually is, but was obscured by a bug in sequelize](https://github.com/sequelize/cli/issues/907).

Lets do the following:

1. Update our postgres URL strings to say `postgres://` isntead of `postgresql://`
2. Update our test script to `db:drop` and `db:create`

We really ought to actually test migration rollback as well, but that's out of scope for this issue.

slifty:
## Description
This PR adds the database concept of TwitterAccountLists, which replace the previous constant-driven twitter account list architecture.

It updates the twitter account list sync logic to use the database.

It is a draft because it does not yet update the newsletter generation to reflect these changes.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits
- [x] Relevant labels assigned

## Steps to Test
1. Update your environment variables so that the DB urls start with `postgres://` instead of `postgresql://`
2. Wipe out your local DB if I'm being honest
3. re-run migrate
4. Add two rows to `twitter_account_lists`: 

```
0 | national      | 1gLkx2LK3yhS-glpsktWYNqBc9H2zsd7C6TvmgWjL5Kg | 2020-06-11 11:58:45.171008-04 | 2020-06-11 11:58:45.171008-04
  1 | northCarolina | 1Z7grw_GQLNMtSVvkUtbXatly_JId7h3yVDX9BWyJetg | 2020-06-11 11:59:13.735023-04 | 2020-06-11 11:59:13.735023-04
```

5. Try running the scraper and all that jazz as you normally would.

## Deploy Notes
- `yarn migrate`
- Update .env variables so that the DB urls start with `postgres://` instead of `postgresql://`
- Add two rows to `twitter_account_lists` (see previous section with data)

## Related Issues
Resolves #364 
Resolves #368

### DIFF
--- a/src/server/migrations/20200531192735-refactor-twitter-account-list-association.js
+++ b/src/server/migrations/20200531192735-refactor-twitter-account-list-association.js
@@ -1,0 +1,81 @@
+import { runPromiseSequence } from '../utils'
+
+module.exports = {
+  up: (queryInterface, Sequelize) => queryInterface.sequelize
+    .transaction(transaction => runPromiseSequence([
+      () => queryInterface.addColumn(
+        'twitter_accounts',
+        'twitter_account_list_id',
+        {
+          type: Sequelize.INTEGER,
+          references: {
+            model: 'twitter_account_lists',
+            key: 'id',
+          },
+        },
+        { transaction },
+      ),
+      () => queryInterface.sequelize.query(
+        `
+        UPDATE twitter_accounts
+        SET    twitter_account_list_id = (
+          SELECT twitter_account_lists.id
+          FROM   twitter_account_lists
+          WHERE  twitter_account_lists.name = twitter_accounts.list_name
+          LIMIT  1
+        )`,
+        { transaction },
+      ),
+      () => queryInterface.removeColumn(
+        'twitter_accounts',
+        'list_name',
+        { transaction },
+      ),
+      () => queryInterface.addConstraint(
+        'twitter_accounts',
+        ['screen_name', 'twitter_account_list_id'],
+        {
+          type: 'unique',
+          name: 'screen_name_twitter_account_list_id',
+          transaction,
+        },
+      ),
+    ])),
+
+  down: (queryInterface, Sequelize) => queryInterface.sequelize
+    .transaction(transaction => runPromiseSequence([
+      () => queryInterface.addColumn(
+        'twitter_accounts',
+        'list_name',
+        {
+          type: Sequelize.STRING(128),
+        },
+        { transaction },
+      ),
+      () => queryInterface.addConstraint(
+        'twitter_accounts',
+        ['screen_name', 'list_name'],
+        {
+          type: 'unique',
+          name: 'screen_name_list_name',
+          transaction,
+        },
+      ),
+      () => queryInterface.sequelize.query(
+        `
+        UPDATE twitter_accounts
+        SET    list_name = (
+          SELECT twitter_account_lists.name
+          FROM   twitter_account_lists
+          WHERE  twitter_account_lists.id = twitter_accounts.twitter_account_list_id
+          LIMIT  1
+        )`,
+        { transaction },
+      ),
+      () => queryInterface.removeColumn(
+        'twitter_accounts',
+        'twitter_account_list_id',
+        { transaction },
+      ),
+    ])),
+}
--- a/src/server/models/TwitterAccount.js
+++ b/src/server/models/TwitterAccount.js
@@ -7,12 +7,12 @@ module.exports = (sequelize, DataTypes) => {
       unique: 'twitterAccountScreenNameListName',
     },
     preferredDisplayName: DataTypes.STRING(512),
-    listName: {
-      type: DataTypes.STRING(128),
-      unique: 'twitterAccountScreenNameListName',
-    },
     isActive: DataTypes.BOOLEAN,
   }, {})
+
+  TwitterAccount.associate = (models) => {
+    TwitterAccount.belongsTo(models.TwitterAccountList)
+  }
 
   TwitterAccount.getByScreenNames = async screenNames => (TwitterAccount.findAll({
     where: {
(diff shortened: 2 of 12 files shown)


src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
```diff
@@ -52,26 +52,26 @@ const extractTwitterAccountsFromRows = (rows, screenNameColumns) => rows
...
+  logger.info(`Syncing twitter accounts for "${twitterAccountList.name}" from Google doc: ${twitterAccountList.googleDocId}`)
+  const scraper = new GoogleSpreadsheetScraper(twitterAccountList.googleDocId, true)
   const rows = await scraper.run()
   const screenNameColumns = extractTwitterScreenNameColumnsFromRow(rows[0] || {})
   const twitterAccounts = extractTwitterAccountsFromRows(rows, screenNameColumns)
   const decoratedTwitterAccounts = decorateObjects(
     twitterAccounts,
     {
-      listName,
+      TwitterAccountListId: twitterAccountList.id, // TODO: this should be lower case
```
reefdog, line 68: Is this TODO backed by an issue? If not, can it be?

src/server/models/TwitterAccount.js
```diff
@@ -31,32 +31,40 @@ module.exports = (sequelize, DataTypes) => {
...
-  TwitterAccount.getActiveByListNames = async listNames => (
-    TwitterAccount.findAll({
-      where: {
-        listName: {
-          [Sequelize.Op.in]: listNames,
+  TwitterAccount.getActiveByListNames = async listNames => TwitterAccount.findAll({
+    where: {
+      isActive: true,
+    },
+    include: [
```
reefdog, line 38: I'm not positive this is doing what you think it's doing; is it actually scoping the accounts return by the list names provided? If you're sure, 👍, but that's not how I read the `include` option, which I thought just literally included the associated model as a child object! (I'm trying to do a code-review-only and not actually run this code, so, leave it up to you to inspect/prove this point!)

src/server/migrations/20200531192735-refactor-twitter-account-list-association.js
```diff
@@ -0,0 +1,59 @@
...
+            key: 'id',
+          },
+        },
+        { transaction },
+      ),
+      () => queryInterface.removeColumn(
+        'twitter_accounts',
+        'list_name',
+        { transaction },
+      ),
```
reefdog, line 22: 🤔 Are we not going to try and migrate existing data? (This doesn't try to map old list names to new `TwitterAccountList` associations.) I can see why not, since we can just run a new sync from Google Sheets. But I wonder if it's even worth leaving the data in this halfway state? I mean after this migration is run but until the next sync from Google, the accounts are basically orphaned from their lists, right? Worth fighting?

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
slifty: It's blocked by the entire model of Sequelize.

In seriousness, yeah I'll make one.  We need to update our models to use camelCase names instead of UpperCase names at definition.

reefdog: _backed_ not _blocked_

src/server/models/TwitterAccount.js
reefdog: Turns out this is how Sequelize wants you to do associative wheres now; there may be a way to exclude the associated model data from the resulting dataset, and that is probably worth two seconds' examination because we don't actually want the data. But! This works!

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
reefdog: Do you mind creating a tracking issue here in our repo to track Sequelize fixing that, and noting the issue number in the TODO comment?

slifty: what about baked?

reefdog: *borked

src/server/migrations/20200531192735-refactor-twitter-account-list-association.js
slifty: I will add this! ty

src/server/models/TwitterAccount.js
slifty: Just an update here: it is weird but we went through it and, indeed, this seems to be the official syntax (so says [StackOverflow](https://stackoverflow.com/questions/18838433/sequelize-find-based-on-association))

gonna mark this as resolved.

reefdog: Wait I thought I marked it as resolved.

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
slifty: BIXED

src/server/migrations/20200531192735-refactor-twitter-account-list-association.js
slifty: FIXED (would benefit from brief re-review just to test this one thing)

slifty:
this took a LOT OF EFFORT TO GET HERE THANK YOU JUSTIN!
... and there's still another PR after this one RIP...

Tomorrow.  I think I'm winded for today.

src/server/queues/twitterAccountListScraperQueue/twitterAccountListScraperJobProcessor.js
reefdog: Bank you.

